PR #4, opened 2026-02-13
feat(helm): v0.3.0

inyellowbus:
appVersion: `v0.1.2-mainnet`

Huge thanks to [Kiln](https://www.kiln.fi/) ([@kilnfi](https://github.com/kilnfi)) for the detailed feedback that shaped this release. You are the best <3

### Added

- `extraInitContainers` — inject custom init containers before the node starts
- `extraContainers` — run sidecar containers alongside the node
- `extraVolumes` and `extraVolumeMounts` — mount additional volumes into the pod and main container
- `podAnnotations` and `podLabels` — custom pod metadata for Vault injection, service mesh, cost tracking, etc.
- `extraEnv` — environment variables for the main container (Downward API, ConfigMap/Secret refs)
- `extraEnvFrom` — inject all keys from a Secret or ConfigMap as environment variables
- `serviceAccount` — create and bind a dedicated ServiceAccount (for Vault auth, cloud IAM)
- `hostPort` — expose ADNL on the host IP without full `hostNetwork` access
- `networkPolicy` — optional NetworkPolicy with public ADNL ingress and configurable TCP CIDRs

### Changed

- `serviceAccount.create` renamed to `serviceAccount.enabled` for consistency with other feature toggles

### Fixed

- Missing logger targets in documentation (`overlay_broadcast`, `adnl_query`, `validate_reject`, `catchain_network`, `block`)
- Added note about HTTP request logging not being available


helm/ton-rust-node/docs/networking.md
```diff
@@ -76,6 +77,40 @@ With `externalTrafficPolicy: Local` the traffic is only routed to pods on the no
 - The `adnl_node.ip_address` in the node config must be set to the node's external IP and the NodePort (not the container port).
 - You need to ensure the pod is scheduled on the node whose IP you configured — use `nodeSelector` or `nodeAffinity`.
 
+## hostPort
+
+Binds only the ADNL UDP port to the host's network interface. The pod stays in the pod network — only the specified port is exposed on the host IP. A lighter alternative to `hostNetwork` that preserves network policy enforcement.
```
evgenuysudlenkovrsquad: Why is hostPort supported only for the ADNL port? If liteserver/jsonRpc are exposed externally anyway, I’m not sure this materially changes the attack surface. The current wording about “minimal attack surface” seems misleading.

helm/ton-rust-node/templates/statefulset.yaml
```diff
@@ -90,12 +102,19 @@ spec:
           command: ["node", "-c", "/main"]
           {{- end }}
           {{- end }}
+          {{- with .Values.extraEnv }}
+          env:
```
rayanebel: You can also add `extraEnvFrom` if we need to inject environment variables from a Kubernetes Secret.

PR #80, opened 2026-04-09
feat(voting): add nodectl vote command for config proposals

Keshoid:
## Summary

- Add `nodectl vote` CLI command for viewing and voting on config proposals
- Improve voting UX: fix hash case sensitivity, better table formatting with cyan headers
- Treat "not a validator" as a warning instead of error in service mode

Closes SMA-68


src/node-control/commands/src/commands/nodectl/vote_cmd.rs
```diff
@@ -0,0 +1,471 @@
...
+ * This software is provided "AS IS", WITHOUT WARRANTY OF ANY KIND.
+ */
+use crate::commands::nodectl::{output_format::OutputFormat, utils::save_config};
+use anyhow::Context;
+use base64::Engine;
+use colored::Colorize;
+use common::app_config::{AppConfig, VotingConfig};
+use contracts::{ConfigContractImpl, ConfigContractWrapper, ConfigProposal, contract_provider};
+use std::{
+    io::{IsTerminal, Write},
```
mrnkslv: mb is it better to add here "stdin/stdout"?  instead of using 
` std::io::stdout().flush()?;` or
    `std::io::stdin().read_line(&mut input)?;`

Keshoid: refactored

mrnkslv (approved):

ITBear (approved):

PR #125, opened 2026-04-27
feat: Configure auto-deploy and auto-topup features

mrnkslv:
feat(nodectl): configurable contracts automation (auto-deploy / topup)
- Add contracts_automation to AppConfig (amounts, tick, toggles, SNP vs TONCore pool deploy)
- Wire contracts monitor to live config; respect auto_deploy / auto_topup
- Expose GET/POST /v1/contracts-automation/settings (partial update, validate, persist)
- Add `nodectl config contracts-automation ls|set` (API-only; amounts in TON on CLI)
- Document in docs/contracts-automation.md, README, and nodectl-setup
- fix: align elections bindings table rule length with column widths


src/node-control/docs/contracts-automation.md
```diff
@@ -0,0 +1,104 @@
...
+"contracts_automation": {
+  "tick_interval_sec": 40,
+  "auto_deploy": true,
+  "auto_topup": true,
+  "wallet_deploy_nanotons": 1100000000,
+  "pool_deploy_nanotons": {
+    "single_nominator": 1100000000,
+    "ton_core": 1100000000
+  },
+  "wallet_topup_nanotons": 10000000000000,
```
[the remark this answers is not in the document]
mrnkslv: done

src/node-control/service/src/contracts/contracts_task.rs
```diff
@@ -55,20 +53,23 @@ struct ContractsMonitor {
...
         let mut cancel = cancellation_ctx.subscribe();
-        let mut interval = tokio::time::interval(tick_interval);
+        // First wait is zero so the first `run` matches `Interval::tick()` semantics.
+        let mut sleep_dur = Duration::from_secs(0);
         loop {
             tokio::select! {
-                _ = interval.tick() => {
+                _ = tokio::time::sleep(sleep_dur) => {
+                    sleep_dur =
+                        Duration::from_secs(self.automation().tick_interval_sec.max(1));
```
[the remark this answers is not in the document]
mrnkslv: done

src/node-control/common/src/app_config.rs
```diff
@@ -734,6 +844,8 @@ pub struct AppConfig {
     /// Default interval for all tasks in seconds
     #[serde(default = "default_tick_interval")]
     pub tick_interval: u64,
+    #[serde(default)]
+    pub contracts_automation: ContractsAutomationConfig,
```
Keshoid: Let's make the name shorter -> `automation`

src/node-control/service/src/http/config_handlers.rs
```diff
@@ -824,6 +873,101 @@ fn build_binding_election_status(
...
+pub async fn v1_contracts_automation_settings_update_handler(
+    state: axum::extract::State<AppState>,
+    req: axum::Json<ContractsAutomationSettingsUpdateRequest>,
+) -> Result<axum::Json<ContractsAutomationSettingsResponse>, AppError> {
+    let req = req.0;
+    if !req.any_field_set() {
+        return Err(AppError::bad_request("at least one setting is required"));
+    }
+
+    let cfg = state.runtime_cfg.get();
```
Keshoid: **Lost-update race in the settings POST handler**

The merge is computed against a snapshot taken *before* `update_and_save`, then written back wholesale inside the closure:

```rust
let cfg = state.runtime_cfg.get();
let next = merge_contracts_automation_update(&cfg.contracts_automation, &req);
next.validate().map_err(|e| AppError::bad_request(e.to_string()))?;
drop(cfg);

state.runtime_cfg.update_and_save(|cfg| {
    cfg.contracts_automation = next;
})
```

If two operators (or a script) POST disjoint partial updates concurrently, both read the same baseline and the second writer clobbers the first. Example: A sends `{"auto_deploy": false}`, B sends `{"tick_interval_sec": 60}` — whichever lands second resets the other field back to the baseline value.

The fix is to move the merge *inside* the `update_and_save` closure so the read-modify-write is atomic under whatever lock `RuntimeConfig` uses. The wrinkle is that `validate()` returns `Result`, and the current closure signature is `FnOnce(&mut AppConfig)` with no error path — so doing this properly means either:

1. Changing `update_and_save` to accept a fallible closure (`FnOnce(&mut AppConfig) -> Result<()>`), or
2. Pre-validating the request shape (range checks on individual fields) before the closure, and trusting that any baseline that already passed validation + a valid patch stays valid.

The same pattern exists in `v1_elections_settings_update_handler`, so this isn't a regression introduced by this PR. Not a release blocker — single-operator usage is fine — but worth a follow-up that touches both handlers (and the `update_and_save` contract) together.

src/node-control/service/src/contracts/contracts_task.rs
```diff
@@ -84,21 +91,22 @@ impl ContractsMonitor {
...
         }
-        if !self.ensure_pools_deployed(&mut seqno).await? {
+        if !self.ensure_pools_deployed(&auto, &mut seqno).await? {
             return Ok(());
         }
-        if !self.ensure_wallet_balances(&mut seqno).await? {
+        if !self.ensure_wallet_balances(&auto, &mut seqno).await? {
             return Ok(());
         }
         if !self.ensure_pool_validator_sets_updated(&mut seqno).await? {
```
Keshoid: **Decouple `update_validator_set` from the master wallet**

`ensure_pool_validator_sets_updated` (this method) sends opcode 6 via `self.master_wallet`, even though the comment a few lines above explicitly notes that the message "must be sent explicitly (**by anyone**)". The corresponding validator wallet for the pool's node is a more natural sender:

- It's already deployed by the time the pool reaches `state == 2` (staking).
- It's auto-topped up, so `POOL_OP_GAS` (0.5 TON) is comfortably covered.
- It scopes the operation to the node that owns the pool — cleaner ownership semantics.

Looking up the wallet for the binding is straightforward — `self.wallets` is keyed by `node_id`, which we already have in the outer loop.

The bigger payoff: once this path no longer needs the master, the toggles fully describe master usage. We can then short-circuit `ensure_master_deployed` (and the rest of `run()`) when `!auto.auto_deploy && !auto.auto_topup`:

```rust
async fn run(&self) -> anyhow::Result<()> {
    let auto = self.automation();
    if !auto.auto_deploy && !auto.auto_topup {
        // Still run ensure_pool_validator_sets_updated — it now uses per-node wallets.
        return self.ensure_pool_validator_sets_updated(...).await.map(|_| ());
    }
    if !self.ensure_master_deployed(&auto).await? {
        return Ok(());
    }
    // ...
}
```

This lets operators run nodectl in observe-or-elections-only mode on hosts where the master is intentionally absent / unfunded.

Not blocking for v0.5.0 (current behavior is correct, just unnecessarily coupled), but a worthwhile follow-up.

src/node-control/commands/src/commands/nodectl/config_contracts_automation_cmd.rs
```diff
@@ -0,0 +1,242 @@
...
+}
+
+#[derive(clap::Args, Clone)]
+pub struct LsCmd {
+    #[arg(long = "format", default_value = "table", help = "Output format: table or json")]
+    format: OutputFormat,
+}
+
+#[derive(clap::Args, Clone)]
+pub struct SetCmd {
```
Keshoid: **CLI ergonomics: shorter shape, grouped by domain**

The current command is verbose enough that the README examples all use line continuations:

```bash
nodectl config contracts-automation set --tick-interval-sec 60
nodectl config contracts-automation set --wallet-deploy 1.1 --wallet-topup 10 --wallet-threshold 5
nodectl config contracts-automation set --pool-deploy-snp 1.1 --pool-deploy-ton-core 2
nodectl config contracts-automation set --auto-deploy false
nodectl config contracts-automation set --auto-topup true
```

The sibling `config elections` already uses a different shape — sub-subcommands per setting, no monolithic `set`. Applying that style here, dropping the `config-` prefix from the name (it's the only `config <name>-…` form in the tree), and hoisting it to the top level so other settings groups can follow the same cleanup:

```bash
nodectl automation ls
nodectl automation tick 60                      # positional, mirrors `elections tick-interval`
nodectl automation wallet --deploy 1.1 --topup 10 --threshold 5
nodectl automation pool --deploy 1.5                       # same amount for both pool kinds
nodectl automation pool --deploy 1.5 --ton-core 2          # shared default + per-kind override
nodectl automation pool --snp 1.1 --ton-core 2             # explicit per-kind
nodectl automation enable deploy                # auto_deploy = true
nodectl automation enable topup
nodectl automation disable deploy
nodectl automation disable topup
```

What this buys:

- **Shorter on average** — the repeated `set --<long-flag-prefix>-…` boilerplate is gone.
- **Consistent with `elections`** — operators learn one mental model.
- **Domain grouping** — `wallet` and `pool` carry their own context, so `--wallet-*` / `--pool-deploy-*` prefixes drop entirely; `--wallet-threshold` becomes just `--threshold` under `wallet`. This also resolves the CLI/API naming asymmetry (no more `--wallet-threshold` ↔ `wallet_balance_threshold` mismatch).
- **Common case for pool** — `--deploy N` under `automation pool` sets both kinds at once (the usual operator default); `--snp` / `--ton-core` either stand alone or override the shared `--deploy`. Validation: at least one of `--deploy`, `--snp`, `--ton-core` must be present.
- **Toggles read as verbs** — `enable deploy` / `disable topup` is more natural than `--auto-deploy true` / `--auto-topup false`.
- **Smaller request bodies** — each invocation sets one domain at a time, which incidentally reduces the surface for the partial-update race flagged in `config_handlers.rs` (concurrent edits touching different domains conflict less often).

The REST API stays exactly as-is — flat partial PATCH with `pool_deploy.single_nominator`, etc. Only the CLI translation layer changes: `--deploy N` (without overrides) expands to `pool_deploy: { single_nominator: N, ton_core: N }`; with overrides, the override wins for that kind.

Not blocking — this is a CLI-surface change, ideally landed before v0.5.0 ships so we don't have to carry a deprecation alias.

src/node-control/common/src/app_config.rs
```diff
@@ -715,6 +715,116 @@ impl Default for LogConfig {
...
+            pool_deploy: PoolDeployAmounts::default(),
+            wallet_topup: default_contracts_wallet_topup(),
+            wallet_balance_threshold: default_contracts_wallet_balance_threshold(),
+        }
+    }
+}
+
+impl ContractsAutomationConfig {
+    /// Validates contracts automation settings (amounts, tick interval).
+    pub fn validate(&self) -> anyhow::Result<()> {
```
Keshoid, line 857: **`validate()` has no upper bounds on monetary fields**

The amount checks currently only reject zero:

```rust
if self.wallet_deploy == 0 {
    anyhow::bail!("contracts_automation.wallet_deploy must be > 0");
}
// ...same shape for pool_deploy.{single_nominator,ton_core}, wallet_topup, wallet_balance_threshold
```

A `u64` lets through values up to ~18 billion TON. The master-balance check at send time would catch a runaway amount before TONs are actually moved, but until then the config sits accepted, and an operator who fat-fingered `--wallet-deploy 1100` (meaning 1.1) wouldn't find out until the contracts monitor tries the next deploy and fails with "insufficient master balance".

A generous upper sanity bound — say 100 000 TON (1e14 nanotons) per field — costs nothing and catches the decimal-place class of mistakes immediately at the CLI/API boundary. Easy to extend:

```rust
const MAX_AMOUNT_NANOTONS: u64 = 100_000 * 1_000_000_000; // 100 000 TON

fn check_amount(name: &str, value: u64) -> anyhow::Result<()> {
    if value == 0 {
        anyhow::bail!("contracts_automation.{name} must be > 0");
    }
    if value > MAX_AMOUNT_NANOTONS {
        anyhow::bail!(
            "contracts_automation.{name} must be <= {} nanotons ({} TON), got {}",
            MAX_AMOUNT_NANOTONS, MAX_AMOUNT_NANOTONS / 1_000_000_000, value
        );
    }
    Ok(())
}
```

`wallet_balance_threshold` doesn't move funds, so its ceiling is more about catching nonsense than safety — same bound is fine, or a higher one if you prefer to be lenient.

Optional add: tighten the zero check on `wallet_deploy` to `>= WALLET_GAS` (0.1 TON), since values below that would let `ensure_master_deployed`'s balance gate pass while the actual deploy txn can't afford gas. Strictly belt-and-braces — the failure mode is "deploy fails on chain", not data loss — but cheap to add while you're in the file.

Not blocking. Mostly UX.

src/node-control/common/src/app_config.rs
```diff
@@ -715,6 +715,116 @@ impl Default for LogConfig {
...
+#[cfg_attr(feature = "openapi", derive(utoipa::ToSchema))]
+pub struct ContractsAutomationConfig {
+    /// Poll interval for the contracts monitor in seconds.
+    #[serde(default = "default_contracts_automation_tick_interval_sec")]
+    pub tick_interval_sec: u64,
+    #[serde(default = "default_contracts_automation_enabled")]
+    pub auto_deploy: bool,
+    #[serde(default = "default_contracts_automation_enabled")]
+    pub auto_topup: bool,
+    #[serde(default = "default_contracts_wallet_deploy", alias = "wallet_deploy_nanotons")]
```
Keshoid: Why do we need alias here?

src/node-control/common/src/app_config.rs
```diff
@@ -715,6 +715,116 @@ impl Default for LogConfig {
...
+            single_nominator: default_contracts_wallet_deploy(),
+            ton_core: default_contracts_wallet_deploy(),
+        }
+    }
+}
+
+/// Parameters for the contracts monitor (auto-deploy wallets/pools, auto-topup validator wallets).
+#[derive(serde::Serialize, serde::Deserialize, Clone, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "openapi", derive(utoipa::ToSchema))]
+pub struct ContractsAutomationConfig {
```
Keshoid, line 787: **Group fields by object — drop the repeated prefixes**

The current shape leans on field-name prefixes to encode grouping:

```rust
pub struct ContractsAutomationConfig {
    pub tick_interval_sec: u64,
    pub auto_deploy: bool,
    pub auto_topup: bool,
    pub wallet_deploy: u64,
    pub pool_deploy: PoolDeployAmounts,
    pub wallet_topup: u64,
    pub wallet_balance_threshold: u64,
}

pub struct PoolDeployAmounts {
    pub single_nominator: u64,
    pub ton_core: u64,
}
```

That makes `wallet_balance_threshold` and `pool_deploy.single_nominator` longer than they need to be, and it's the root cause of the CLI/API asymmetry (`--wallet-threshold` ↔ `wallet_balance_threshold`, `--pool-deploy-snp` ↔ `single_nominator`). Lift the grouping into nested structs and the prefixes fall away:

```rust
pub struct ContractsAutomationConfig {
    pub tick_interval_sec: u64,
    pub auto_deploy: bool,
    pub auto_topup: bool,
    pub wallet: WalletAmounts,
    pub pool: PoolAmounts,
}

pub struct WalletAmounts {
    pub deploy: u64,
    pub topup: u64,
    pub threshold: u64,
}

pub struct PoolAmounts {
    pub snp: u64,
    pub ton_core: u64,
}
```

JSON:

```json
{
  "tick_interval_sec": 40,
  "auto_deploy": true,
  "auto_topup": true,
  "wallet": { "deploy": 1100000000, "topup": 10000000000, "threshold": 5000000000 },
  "pool":   { "snp": 1100000000,    "ton_core": 1100000000 }
}
```

Pairs cleanly with the CLI shape suggested in the other thread — API field names match flag names 1:1:

- `automation wallet --deploy 1.1 --topup 10 --threshold 5` → PATCH `{ "wallet": { ... } }`
- `automation pool --snp 1.1 --ton-core 2` → PATCH `{ "pool": { ... } }`

Decisions made along the way:

- **Singular `wallet`/`pool`** — these are settings sections, not collections of instances. `AppConfig` already uses plural `wallets`/`pools` for the actual maps; reusing those names here would be confusing.
- **`snp` vs `single_nominator`** — `snp` is consistent with `PoolKind::SNP` and the agreed `--snp` flag. The spelled-out form is just longer to read and write.
- **`ton_core` (not `toncore`)** — keeps parity with `PoolKind::TONCore` (two words conceptually). Clap auto-converts to `--ton-core`.
- **Drop the `*_nanotons` aliases** — they exist now only as defensive backwards-compat with what *would have been* the v0.4 shape, but this feature has not shipped yet (it's still in `release/nodectl/v0.5.0`). No on-disk config in the wild uses them. Cleaner to land the new schema with nothing legacy to carry.

Worth doing in this PR rather than later — once v0.5.0 ships, renaming becomes a migration problem.

src/node-control/docs/contracts-automation.md
```diff
@@ -0,0 +1,104 @@
+# Contracts automation (auto-deploy / auto-topup)
+
+The **contracts monitor** (part of the nodectl service) periodically:
```
Keshoid: Let's use `contracts task` instead of monitor. `Monitor` will be reserved for monitoring commands.

src/node-control/service/src/http/http_server_task.rs
```diff
@@ -161,6 +161,10 @@ pub(crate) fn routes(enable_swagger: bool, state: AppState) -> axum::Router {
         .route("/v1/wallets", axum::routing::get(v1_wallets_handler))
         .route("/v1/pools", axum::routing::get(v1_pools_handler))
         .route("/v1/bindings", axum::routing::get(v1_bindings_handler))
+        .route(
+            "/v1/contracts-automation/settings",
```
Keshoid: Rename to `automation`

src/node-control/service/src/http/http_server_task.rs
```diff
@@ -176,6 +180,12 @@ pub(crate) fn routes(enable_swagger: bool, state: AppState) -> axum::Router {
             "/v1/elections/settings",
             axum::routing::post(super::config_handlers::v1_elections_settings_update_handler),
         )
+        .route(
+            "/v1/contracts-automation/settings",
```
Keshoid: Rename to `automation`

Keshoid (changes requested):

src/node-control/common/src/app_config.rs

src/node-control/service/src/http/http_server_task.rs
mrnkslv: done

src/node-control/service/src/http/http_server_task.rs
mrnkslv: done

src/node-control/common/src/app_config.rs
mrnkslv: the first naming was like alias. But now we don't need it anymore.deleted

src/node-control/docs/contracts-automation.md
mrnkslv: ok

src/node-control/commands/src/commands/nodectl/config_contracts_automation_cmd.rs
mrnkslv: done

src/node-control/common/src/app_config.rs
mrnkslv: done

src/node-control/common/src/app_config.rs
mrnkslv: changed

src/node-control/service/src/contracts/contracts_task.rs
mrnkslv: done

src/node-control/service/src/contracts/contracts_task.rs
```diff
@@ -435,6 +488,22 @@ impl ContractsMonitor {
...
         {
+            let validator_wallet = match self.wallets.get(node_id.as_str()) {
+                Some(w) => (*w).clone(),
+                None => {
+                    tracing::warn!(
+                        target: "contracts",
+                        "[{}] ensure_pool_validator_sets_updated: no validator wallet in config (skip TonCore pools)",
+                        node_id
+                    );
+                    continue;
```
Keshoid, line 500: When the validator wallet is missing we `continue` without flipping `all_updated = false`. That means the function can still return `true` while we silently skipped every TonCore pool for that node, and the auto-mode branch in `run()` will log `"all contracts are ready"` even though work was left undone.

Please set `all_updated = false;` before `continue;` so the readiness log stays honest (and the auto-mode loop retries on the next tick).

src/node-control/service/src/contracts/contracts_task.rs
```diff
@@ -435,6 +488,22 @@ impl ContractsMonitor {
...
+                }
+            };
+
+            let wallet_addr = validator_wallet.address().await?;
+            let mut seqno =
+                provider.get_method(wallet_addr.to_string(), "seqno", vec![]).await?.i64(0)?;
```
Keshoid: The validator wallet's `seqno` is fetched here once per node-with-TonCore-binding **per tick**, before we know whether any pool actually needs `update_validator_set`. In steady state (pool `state != 2` or `vsc_count >= 2`) none of them do, so this is a redundant RPC every tick on every such node.

Please make the fetch lazy — initialize it only on the first send inside the inner loop. Something like:

```rust
let mut seqno: Option<i64> = None;

for pool in pool_binding.inner_pools() {
    // ... existing checks; `continue` early as today ...

    let current_seqno = match seqno {
        Some(s) => s,
        None => {
            let s = provider
                .get_method(wallet_addr.to_string(), "seqno", vec![])
                .await?
                .i64(0)?;
            seqno = Some(s);
            s
        }
    };

    // ... build_message with Some(u32::try_from(current_seqno)?) ...

    seqno = Some(current_seqno + 1);
}
```

This drops the RPC count to zero on quiet ticks (the common case).

src/node-control/service/src/runtime_config.rs
```diff
@@ -513,7 +559,16 @@ impl RuntimeConfig for RuntimeConfigStore {
     }
 
     fn update_and_save(&self, f: Box<dyn FnOnce(&mut AppConfig) + Send>) -> anyhow::Result<()> {
```
Keshoid, line 576: This trait impl duplicates the body of the inherent `RuntimeConfigStore::update_and_save` method verbatim (same `try_update_and_save` wrap + same three-arm `match` with the same `"unexpected rejection from infallible update_and_save closure"` message). Please collapse it to a one-line forward:

```rust
fn update_and_save(&self, f: Box<dyn FnOnce(&mut AppConfig) + Send>) -> anyhow::Result<()> {
    RuntimeConfigStore::update_and_save(self, f)
}
```

One copy of the rejection-mapping logic is enough.

src/node-control/common/src/app_config.rs
```diff
@@ -734,6 +886,8 @@ pub struct AppConfig {
     /// Default interval for all tasks in seconds
     #[serde(default = "default_tick_interval")]
     pub tick_interval: u64,
+    #[serde(default, alias = "contracts_automation")]
```
Keshoid: Please drop the `alias = "contracts_automation"`. The feature has not shipped yet (still on `release/nodectl/v0.5.0`), so there are no on-disk configs in the wild that use the old root-level key — we don't need to carry the migration. The companion test `app_config_deserializes_legacy_root_key_contracts_automation` should be removed along with it.

src/node-control/service/src/runtime_config.rs
mrnkslv: done

src/node-control/service/src/contracts/contracts_task.rs
mrnkslv: ok

src/node-control/service/src/http/config_handlers.rs
mrnkslv: already done

src/node-control/service/src/contracts/contracts_task.rs
mrnkslv: done

Keshoid (approved):

PR #128, opened 2026-04-30
Feature/sma 55 add cli commands for adaptive staking parameters

mrnkslv:
## Summary
- Adaptive staking CLI — extend nodectl config elections and related service wiring so adaptive staking parameters can be managed from the CLI, with README updates and small app_config adjustments.
- Validator deposit amount — align TONCore deposit-validator with the pool’s fixed 1 TON processing fee (outbound message value = stake + fee), document it, and centralize the fee constant in contract message helpers.
- Undeployed TONCore pools — treat uninitialized / missing contracts as not deployed instead of a generic error, merge slot init params from local config into pool list DTOs when chain data is absent, add --validator-share-percent (human-friendly percents vs basis points), refresh table/JSON display and tests, and document config pool ls behavior.
## Changes
- commands: config_elections_cmd (adaptive staking), service_api_cmd, config_pool_cmd (deposit math, --validator-share-percent, pool list state styling, serde defaults where needed).
- service: config_handlers / config_handlers_tests (TONCore pool slots: RPC + config merge, not deployed state), http_server_task (routes/handlers for adaptive elections API).
- contracts: nominator/messages.rs — fee constant and deposit message value semantics.
- common: app_config.rs — minor tweak tied to elections/config flow.
- README: elections adaptive usage, deposit-validator fee note, TONCore pool add (validator-share-percent) and pool ls (not deployed + config fallbacks).


src/node-control/commands/src/commands/nodectl/config_elections_cmd.rs
```diff
@@ -170,6 +195,8 @@ fn print_elections_settings_table(view: &ElectionsSettingsView) {
     println!("  {:<20} {}", "Stake Policy:".cyan().bold(), view.stake_policy);
     println!("  {:<20} {}", "Max Factor:".cyan().bold(), view.max_factor);
     println!("  {:<20} {}s", "Tick Interval:".cyan().bold(), view.tick_interval);
+    println!("  {:<28} {}", "adaptive_sleep_period_pct:".cyan().bold(), view.sleep_period_pct);
```
Keshoid: **Display alignment mismatch**

The first three labels use `{:<20}` width, but these two switch to `{:<28}`. `"adaptive_sleep_period_pct:"` is 27 chars so the value column starts at a different offset than "Stake Policy", "Max Factor", "Tick Interval" above.

Either widen all labels to the same width, or shorten these (e.g. `"Sleep Period Pct:"` / `"Waiting Period Pct:"`) to fit the existing `{:<20}`.

src/node-control/service/src/http/config_handlers.rs
```diff
@@ -662,6 +673,26 @@ pub async fn v1_pools_handler(
     Ok(axum::Json(PoolsResponse { ok: true, result: views }))
 }
 
+/// When on-chain `get_pool_data` is unavailable, copy deploy parameters from local config
+/// into the DTO so `GET /v1/pools` still lists validator share / stake thresholds.
+fn merge_toncore_slot_config_fallbacks(dto: &mut TonCorePoolSlotDto, slot_cfg: &TonCorePoolConfig) {
```
Keshoid, line 846: **Fallback overwrites real unknown state on active accounts**

`merge_toncore_slot_config_fallbacks` runs unconditionally at the end of `fetch_toncore_slot_dto` — including when the account is `"active"` but `get_pool_data` failed (the `tracing::warn` branch). In that case the DTO fields are still `None` and this fallback fills them from local config, which may now differ from on-chain reality.

The old code correctly left those fields empty on `get_pool_data` failure. Consider only merging fallbacks when `state != "active"`, or adding a marker that the data comes from local config.

src/node-control/commands/src/commands/nodectl/config_elections_cmd.rs
```diff
@@ -233,10 +260,49 @@ struct ElectionsSettingsBody {
...
+    ec.sleep_period_pct = sleep;
+    ec.waiting_period_pct = waiting;
+    ec.validate_timing_fields()
+}
+
+fn validate_adaptive_pct_range(v: f64) -> anyhow::Result<()> {
+    if !(0.0..=1.0).contains(&v) {
+        anyhow::bail!("value must be in range [0.0, 1.0]");
+    }
+    Ok(())
```
Keshoid: **Redundant range check**

`validate_adaptive_pct_range` checks `[0.0, 1.0]`, then `validate_sleep_waiting_pair` -> `ElectionsConfig::validate_timing_fields()` checks the exact same range again. This standalone check is dead validation — the pair validation already covers it. Consider removing it or keeping only one path.

src/node-control/commands/src/commands/nodectl/config_elections_cmd.rs
```diff
@@ -233,10 +260,49 @@ struct ElectionsSettingsBody {
...
+        .and_then(|v| v.as_f64())
+        .ok_or_else(|| anyhow::anyhow!("elections settings: missing waiting_period_pct"))?;
+    Ok((sleep, waiting))
+}
+
+fn validate_sleep_waiting_pair(sleep: f64, waiting: f64) -> anyhow::Result<()> {
+    let mut ec = ElectionsConfig::default();
+    ec.sleep_period_pct = sleep;
+    ec.waiting_period_pct = waiting;
+    ec.validate_timing_fields()
```
Keshoid: **Unnecessary allocation**

`ElectionsConfig::default()` allocates a `HashMap`, `String`-keyed map, etc. just to validate two floats. Since you only need `sleep <= waiting && both in [0,1]`, a plain free function taking `(f64, f64)` (or inlining the three comparisons) would be cleaner and avoid the allocation.

src/node-control/commands/src/commands/nodectl/config_pool_cmd.rs
```diff
@@ -528,6 +582,23 @@ fn display_state(state: &i32) -> ColoredString {
     }
 }
 
+/// Row "State" for TONCore table: account / deployment state wins for undeployed accounts; then
+/// on-chain pool lifecycle from `pool_state` when present.
+fn display_toncore_slot_row(s: &TonCorePoolSlotView) -> ColoredString {
+    if matches!(s.state.as_str(), "not deployed" | "uninit") {
```
Keshoid: **Dead branch — `uninit` will never arrive here**

The API side now maps `AccountState::Uninitialized` to `"not deployed"` before serializing the DTO, so `s.state` will never be `"uninit"` when this code runs. The `"uninit"` arm in the match is dead code. Consider removing it.

src/node-control/commands/src/commands/nodectl/config_elections_cmd.rs
```diff
@@ -34,6 +34,10 @@ pub enum ElectionsAction {
     TickInterval(TickIntervalCmd),
     /// Set the max-factor
     MaxFactor(MaxFactorCmd),
+    /// Set AdaptiveSplit50 minimum wait fraction (`sleep_period_pct` in config, 0.0–1.0, must be ≤ waiting period)
+    AdaptiveSleepPeriodPct(AdaptiveSleepPeriodPctCmd),
```
Keshoid: **Command name is too long**

`config elections adaptive-sleep-period-pct` / `adaptive-waiting-period-pct` — this is 44/46 chars just for the subcommand path. Painful to type and hard to remember.

Suggestions:
- Shorter long form: `sleep-pct` / `wait-pct` (or `adaptive-sleep` / `adaptive-wait`)
- Add a short alias via `#[command(alias = "sleep-pct")]` or similar so power users don't have to spell out the full name every time

Keshoid (commented):
PR review — 6 issues found (details in inline comments).

Keshoid (changes requested):

src/node-control/commands/src/commands/nodectl/config_elections_cmd.rs
mrnkslv: done

src/node-control/service/src/http/config_handlers.rs
mrnkslv: done

src/node-control/commands/src/commands/nodectl/config_elections_cmd.rs
mrnkslv: done

src/node-control/commands/src/commands/nodectl/config_elections_cmd.rs
mrnkslv: done

src/node-control/commands/src/commands/nodectl/config_pool_cmd.rs
mrnkslv: ready

Keshoid (commented):
Review of code quality / Rust idiom. Most are cleanup; (1) and (3) are the two I'd want addressed before merge.

src/node-control/commands/src/commands/nodectl/config_elections_cmd.rs
```diff
@@ -362,6 +429,58 @@ impl MaxFactorCmd {
...
 
+impl AdaptiveSleepPeriodPctCmd {
+    pub async fn run(
+        &self,
+        url: Option<&str>,
+        token: Option<&str>,
+        config_path: Option<&str>,
+    ) -> anyhow::Result<()> {
+        let base_url = resolve_service_url(url, config_path)?;
+        let (_cur_sleep, cur_waiting) = fetch_adaptive_timing_percentages(&base_url, token).await?;
```
Keshoid: **Design — collapse to one atomic subcommand.**

This `GET → validate → POST` is racy (another writer can shift the other field between the two calls) and forces operators into a strict order: going `(0.2, 0.4) → (0.6, 0.8)` requires raising `waiting` first. The server already accepts both fields in one POST and runs `validate_timing_fields` on the merged pair (`config_handlers.rs:1537-1549`), so the client check is mostly redundant.

Replace `AdaptiveSleepPeriodPctCmd` + `AdaptiveWaitingPeriodPctCmd` with a single subcommand:

```rust
/// Set AdaptiveSplit50 timing fractions (atomic)
AdaptiveTiming(AdaptiveTimingCmd)

struct AdaptiveTimingCmd {
    #[arg(long)] sleep: Option<f64>,
    #[arg(long)] waiting: Option<f64>,
}
```

This drops `fetch_adaptive_timing_percentages`, `validate_sleep_waiting_pair`, and `~80` lines of near-duplicate `impl ... Cmd` blocks.

src/node-control/commands/src/commands/nodectl/config_elections_cmd.rs
```diff
@@ -167,9 +194,11 @@ struct BindingElectionView {
 
 fn print_elections_settings_table(view: &ElectionsSettingsView) {
     println!("\n{} {}\n", "OK".green().bold(), "Elections Configuration".green());
-    println!("  {:<20} {}", "Stake Policy:".cyan().bold(), view.stake_policy);
-    println!("  {:<20} {}", "Max Factor:".cyan().bold(), view.max_factor);
-    println!("  {:<20} {}s", "Tick Interval:".cyan().bold(), view.tick_interval);
+    println!("  {:<28} {}", "Stake Policy:".cyan().bold(), view.stake_policy);
+    println!("  {:<28} {}", "Max Factor:".cyan().bold(), view.max_factor);
+    println!("  {:<28} {}s", "Tick Interval:".cyan().bold(), view.tick_interval);
+    println!("  {:<28} {}", "adaptive_sleep_period_pct:".cyan().bold(), view.sleep_period_pct);
```
Keshoid: Label uses the raw config key (`adaptive_sleep_period_pct:`) while every other row uses human labels (`Stake Policy:`, `Max Factor:`). Match the style:

```rust
println!("  {:<28} {}", "Adaptive sleep fraction:".cyan().bold(), view.sleep_period_pct);
println!("  {:<28} {}", "Adaptive wait fraction:".cyan().bold(), view.waiting_period_pct);
```

src/node-control/service/src/http/config_handlers.rs
```diff
@@ -675,72 +706,90 @@ pub async fn v1_pools_handler(
...
+                slot: slot_name.to_string(),
+                address: address_str,
+                state: "error".to_string(),
+                ..Default::default()
+            },
+        },
     };
 
-    // Only query pool data when the contract is active
+    let merge_fallback = dto.state != "active";
```
Keshoid: **Active + RPC failure now leaves the DTO empty.**

`merge_fallback = dto.state != "active"`, so when `state == "active"` and `get_pool_data()` fails (line 778), `validator_share` / `max_nominators` / stake thresholds stay `None`. That's exactly the operator's worst-case: `config pool ls` shows `-` everywhere on a flaky RPC, even though the local config has the values.

Docstring justifies it ("avoid stale local config"), but the table is the operator's truth source. Two options:
- Merge fallbacks unconditionally and add a `data_source: "chain" | "config"` discriminator on the DTO; or
- Keep current behavior but include the slot address in a `warn!` and document that fields will be sparse.

I'd take the first.

src/node-control/service/src/http/config_handlers.rs
```diff
@@ -664,6 +675,26 @@ pub async fn v1_pools_handler(
     Ok(axum::Json(PoolsResponse { ok: true, result: views }))
 }
 
+/// When on-chain `get_pool_data` is unavailable, copy deploy parameters from local config
+/// into the DTO so `GET /v1/pools` still lists validator share / stake thresholds.
+fn merge_toncore_slot_config_fallbacks(dto: &mut TonCorePoolSlotDto, slot_cfg: &TonCorePoolConfig) {
+    let Some(p) = slot_cfg.params.as_ref() else {
+        return;
+    };
+    if dto.validator_share.is_none() {
```
Keshoid: Use `Option::get_or_insert` to drop the cascade:

```rust
fn merge_toncore_slot_config_fallbacks(dto: &mut TonCorePoolSlotDto, cfg: &TonCorePoolConfig) {
    let Some(p) = cfg.params.as_ref() else { return };
    dto.validator_share.get_or_insert(p.validator_share);
    dto.max_nominators.get_or_insert(p.max_nominators);
    dto.min_validator_stake.get_or_insert(p.min_validator_stake);
    dto.min_nominator_stake.get_or_insert(p.min_nominator_stake);
}
```

src/node-control/service/src/http/config_handlers.rs
```diff
@@ -675,72 +706,90 @@ pub async fn v1_pools_handler(
...
 async fn fetch_toncore_slot_dto(
     rpc_client: Arc<ClientJsonRpc>,
     slot_idx: usize,
-    config_address: Option<String>,
+    slot_cfg: TonCorePoolConfig,
     cached: Option<Arc<dyn NominatorWrapper>>,
 ) -> TonCorePoolSlotDto {
     let slot_name = if slot_idx == 0 { "even" } else { "odd" };
-    let address = match &cached {
+    let addr_opt: Option<MsgAddressInt> = match &cached {
```
Keshoid, line 887: `fetch_toncore_slot_dto` double-matches on `addr_opt`: once here to build the initial DTO, then again inside `if dto.state == "active"` (line 760) via `if let Some(addr) = addr_opt`. The second match is only reachable when the first was `Some(addr)`. Flatten by threading the address through:

```rust
let (mut dto, addr_for_active) = match addr_opt {
    None => (dto_not_deployed(...), None),
    Some(addr) => match rpc_client.get_address_information(&addr).await {
        Ok(info) => (dto_from_info(...), matches!(info.state, AccountState::Active).then_some(addr)),
        Err(_)   => (dto_error(...), None),
    },
};
if let Some(addr) = addr_for_active { /* fetch pool data */ }
```

Removes a level of nesting and the `unwrap_or_else` Arc construction.

src/node-control/commands/src/commands/nodectl/config_pool_cmd.rs
```diff
@@ -273,21 +285,60 @@ struct PoolAddCoreBody<'a> {
...
 }
 
+fn validator_share_percent_to_bp(pct: f64) -> anyhow::Result<u16> {
+    if !pct.is_finite() || !(0.0..=100.0).contains(&pct) {
+        anyhow::bail!(
+            "validator_share_percent must be a finite number in [0.0, 100.0] (got {pct})"
+        );
+    }
+    let bp = (pct * 100.0).round();
+    if !(0.0..=10_000.0).contains(&bp) {
```
Keshoid: Dead check. After bounding `pct ∈ [0.0, 100.0]` and `.round()`, `bp` is provably in `[0.0, 10_000.0]` — this `if` can never fire. Simplify:

```rust
fn share_pct_to_bp(pct: f64) -> anyhow::Result<u16> {
    anyhow::ensure!(
        pct.is_finite() && (0.0..=100.0).contains(&pct),
        "validator-share-percent must be finite in [0.0, 100.0] (got {pct})"
    );
    Ok((pct * 100.0).round() as u16)
}
```

Also rename `validator_share_percent_to_bp` → `share_pct_to_bp` (the file is already in pool-share scope).

src/node-control/commands/src/commands/nodectl/config_pool_cmd.rs
```diff
@@ -528,6 +582,23 @@ fn display_state(state: &i32) -> ColoredString {
     }
 }
 
+/// Row "State" for TONCore table: account / deployment state wins for undeployed accounts; then
+/// on-chain pool lifecycle from `pool_state` when present.
+fn display_toncore_slot_row(s: &TonCorePoolSlotView) -> ColoredString {
```
Keshoid, line 643: Stringly-typed state matching across the wire. `s.state` is a `String` from the HTTP DTO; this matches on magic strings (`"not deployed"`, `"error"`, `"active"`, `"frozen"`) that the server produces. The HTTP side already has an `AccountState` enum (`ton-http-api-client/src/v2/data_models.rs:46-58`). Define a CLI-side enum with `#[serde(rename_all = "lowercase")]` (+ `"not deployed"` rename) — catches typos at compile time and the `match` becomes exhaustive.

src/node-control/commands/src/commands/nodectl/config_pool_cmd.rs
```diff
@@ -605,23 +676,32 @@ impl PoolDepositValidatorCmd {
...
                 nanotons_to_tons_f64(wallet_info_data.balance),
+                nanotons_to_tons_f64(msg_value_nanotons.saturating_add(gas_reserve)),
                 self.amount,
+                nanotons_to_tons_f64(pool_messages::DEPOSIT_VALIDATOR_POOL_FEE_NANOTONS),
             );
         }
 
         println!(
-            "\n{}\n  Binding: {}\n  Wallet:  {} ({})\n  Pool:    {}\n  Amount:  {:.9} TON\n",
+            "\n{}\n  Binding: {}\n  Wallet:  {} ({})\n  Pool:    {}\n  Credited to validator stake: {:.9} TON\n  Message value (stake + {:.9} TON pool fee): {:.9} TON\n",
```
Keshoid: This multi-line format string with embedded `\n` is hard to scan. Break into stacked `println!`s — same output, less escaping:

```rust
println!("\n{}", "Deposit validator summary:".cyan().bold());
println!("  Binding: {}", self.binding);
println!("  Wallet:  {} ({})", binding.wallet, wallet_address);
println!("  Pool:    {}", pool_address);
println!("  Credited to validator stake: {:.9} TON", self.amount);
println!(
    "  Message value (stake + {:.9} TON pool fee): {:.9} TON\n",
    nanotons_to_tons_f64(pool_messages::DEPOSIT_VALIDATOR_POOL_FEE_NANOTONS),
    nanotons_to_tons_f64(msg_value_nanotons),
);
```

src/node-control/service/src/http/config_handlers_tests.rs
```diff
@@ -170,7 +170,8 @@ async fn pools_toncore_slot_with_no_address_is_not_deployed() {
     assert_eq!(slots[0]["state"], "not deployed");
```
Keshoid, line 265: Test name (`pools_toncore_slot_with_no_address_is_not_deployed`) is now misleading — it also asserts the config-fallback merge (`validator_share == 4000`). Rename to `..._is_not_deployed_with_config_fallback` or split into two tests.

src/node-control/common/src/app_config.rs
```diff
@@ -508,11 +508,13 @@ fn default_tick_interval() -> u64 {
     40
 }
 
-fn default_waiting_pct() -> f64 {
+/// Default `waiting_period_pct` for [`ElectionsConfig`] and API DTOs that omit the field.
+pub fn default_waiting_pct() -> f64 {
```
Keshoid: These were made `pub` solely so the CLI can use them in `#[serde(default = "common::app_config::default_sleep_pct")]` on `ElectionsSettingsView` (`config_elections_cmd.rs:177-180`). But the API DTO always sends these fields (`config_handlers.rs:849-852`), so the `#[serde(default)]` on the CLI view is dead weight.

Either drop both the CLI fallback and the `pub`, or add a one-line comment on the CLI side explaining the fallback exists for forward-compat with older daemons.

src/node-control/commands/src/commands/nodectl/config_pool_cmd.rs
```diff
@@ -311,8 +362,9 @@ impl PoolAddCoreCmd {
         api_post(&base_url, "/v1/pools/core", token, &body).await?;
 
         let mut info_parts = Vec::new();
-        if let Some(vs) = self.validator_share {
-            info_parts.push(format!("validator_share={vs} bp"));
+        if let Some(vs) = resolved_validator_share {
+            let pct = vs as f64 / 100.0;
+            info_parts.push(format!("validator_share={vs} bp (~{pct:.2}%)"));
```
Keshoid, line 382: Minor: `vs as f64 / 100.0` is the bp→percent inverse of `share_pct_to_bp`. If a third caller appears, extract a `bp_to_pct(u16) -> f64` helper alongside it.

src/node-control/commands/src/commands/nodectl/config_elections_cmd.rs
```diff
@@ -34,6 +34,12 @@ pub enum ElectionsAction {
     TickInterval(TickIntervalCmd),
     /// Set the max-factor
     MaxFactor(MaxFactorCmd),
+    /// Set AdaptiveSplit50 minimum wait fraction (`sleep_period_pct` in config, 0.0–1.0, must be ≤ waiting period)
+    #[command(alias = "sleep-pct", alias = "adaptive-sleep")]
+    AdaptiveSleepPeriodPct(AdaptiveSleepPeriodPctCmd),
+    /// Set AdaptiveSplit50 maximum wait fraction (`waiting_period_pct` in config, 0.0–1.0, must be ≥ sleep period)
+    #[command(alias = "wait-pct", alias = "adaptive-wait")]
+    AdaptiveWaitingPeriodPct(AdaptiveWaitingPeriodPctCmd),
```
Keshoid: **Naming + collapse into a single subcommand.**

`adaptive` describes the stake policy (`AdaptiveSplit50`), but this command sets **wait bounds** (min/max as fractions of election duration). Name it `wait` with `--min` / `--max` — the pair reads as bounds with an obvious `min ≤ max` invariant, no need to explain that `sleep` is the lower one.

```bash
nodectl config elections wait --min 0.15 --max 0.45
nodectl config elections wait --min 0.15            # partial update
nodectl config elections wait --max 0.45            # partial update
```

Clap side:

```rust
/// Set min/max wait fractions for AdaptiveSplit50 timing (atomic)
#[command(alias = "wait-pct")]
Wait(WaitCmd)

#[derive(clap::Args, Clone)]
pub struct WaitCmd {
    /// Minimum wait as fraction of election duration (0.0–1.0)
    #[arg(long)]
    min: Option<f64>,
    /// Maximum wait as fraction of election duration (0.0–1.0)
    #[arg(long)]
    max: Option<f64>,
}
```

Replaces both `AdaptiveSleepPeriodPct` and `AdaptiveWaitingPeriodPct`, removes `fetch_adaptive_timing_percentages` + `validate_sleep_waiting_pair` (server validates the pair atomically), and leaves one alias (`wait-pct`) instead of four. Partial updates still work — the server merges the supplied fields with current state and validates the pair.

Keshoid: Concrete clap help texts (semantics from `adaptive_strategy.rs:40-65` — `--min` is the earliest moment we're willing to stake even with enough peers, `--max` is the latest moment we'll keep waiting for peers before staking regardless).

Worth surfacing in the help: **these timings currently only affect the `AdaptiveSplit50` stake policy** — for `minimum` / `split50` / `fixed` they're stored but unused. Mention that on the subcommand so operators don't expect them to gate non-adaptive policies.

```rust
/// Set the AdaptiveSplit50 staking window — when to start staking and how long to wait for peers.
///
/// Only used when the stake policy is `adaptive_split50`; ignored for `minimum`, `split50`,
/// and `fixed`.
#[command(alias = "wait-pct")]
Wait(WaitCmd)

#[derive(clap::Args, Clone)]
pub struct WaitCmd {
    /// Earliest stake submission, as fraction of election duration
    #[arg(
        long,
        long_help = "Defer staking until this fraction of the election window has elapsed, \
                     even when there are already enough peers. Range [0.0, 1.0]; default 0.2. \
                     Only applied under the adaptive_split50 stake policy."
    )]
    min: Option<f64>,

    /// Latest peer-wait deadline, as fraction of election duration
    #[arg(
        long,
        long_help = "Keep waiting for enough peers until this fraction of the election window. \
                     After this point, stake regardless of peer count. Range [0.0, 1.0]; \
                     must be \u2265 --min; default 0.4. Only applied under the adaptive_split50 \
                     stake policy."
    )]
    max: Option<f64>,
}
```

src/node-control/commands/src/commands/nodectl/config_elections_cmd.rs
mrnkslv: done

src/node-control/commands/src/commands/nodectl/config_elections_cmd.rs
mrnkslv: done

src/node-control/service/src/http/config_handlers_tests.rs
mrnkslv: done

src/node-control/commands/src/commands/nodectl/config_pool_cmd.rs
mrnkslv: done

src/node-control/service/src/http/config_handlers.rs

src/node-control/commands/src/commands/nodectl/config_pool_cmd.rs
mrnkslv: done

src/node-control/commands/src/commands/nodectl/config_pool_cmd.rs
mrnkslv: done

src/node-control/common/src/app_config.rs

src/node-control/commands/src/commands/nodectl/config_pool_cmd.rs
mrnkslv: done

src/node-control/commands/src/commands/nodectl/config_elections_cmd.rs
mrnkslv: done

src/node-control/service/src/http/config_handlers.rs

src/node-control/service/src/http/config_handlers.rs
mrnkslv: done

src/node-control/commands/src/commands/nodectl/config_elections_cmd.rs
mrnkslv: done

Keshoid (commented):
Second pass — three issues plus two minor wording nits. (1) is a real display bug Copilot flagged in the first round but never got addressed.

src/node-control/commands/src/commands/nodectl/config_pool_cmd.rs
```diff
@@ -528,6 +642,24 @@ fn display_state(state: &i32) -> ColoredString {
     }
 }
 
+/// Row "State" for TONCore table: account / deployment state wins for undeployed accounts; then
+/// on-chain pool lifecycle from `pool_state` when present.
+fn display_toncore_slot_row(s: &TonCorePoolSlotView) -> ColoredString {
+    use TonCorePoolSlotWireState::*;
+    if matches!(s.state, NotDeployed) {
+        return "not deployed".yellow();
```
Keshoid, line 646: **`"not deployed"` overflows the `{:<10}` State column.**

This branch returns the literal `"not deployed"` (12 chars), but the table header and data row in `print_toncore_table` both use `{:<10}` for the State column. Any undeployed slot pushes Balance / Noms / Share / Address one tab to the right. Copilot flagged this on 2026-04-30; the column width never moved.

Fix one of:
- widen the State column to `{:<13}` in both the header and data `println!` of `print_toncore_table`, or
- shorten the displayed token here (e.g. `"undeployed"` is 10 chars — note that changing the wire form in `TonCorePoolSlotWireState::as_wire` would also touch the API contract, so the column-width fix is cheaper).

src/node-control/commands/src/commands/nodectl/config_pool_cmd.rs
```diff
@@ -347,8 +461,8 @@ struct TonCorePoolSlotView {
     slot: String,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     address: Option<String>,
-    /// Account state: "active", "uninit", "frozen", "not deployed", or "error".
-    state: String,
+    #[serde(default, deserialize_with = "deserialize_toncore_pool_slot_wire_state")]
+    state: TonCorePoolSlotWireState,
```
Keshoid, line 446: **`data_source` is server-only — half of the option-1 fix landed.**

The `data_source: chain | config` discriminator was the operator-visible half of the unconditional-fallback design (review comment 3227865036). The server now emits it (`config_handlers.rs:107, 819-825`), but `TonCorePoolSlotView` here doesn't deserialize it and `pool ls` doesn't display it. So on a flaky RPC, the operator sees populated `validator_share` / stake thresholds but no signal that those are stale `config.toml` defaults rather than live chain reads — which is the exact ambiguity the discriminator was supposed to remove.

Close the loop one of two ways:
- add a `data_source` field to this struct and append a `Src` column to the TONCore table (or a `*` suffix on share/stake cells when `data_source == Config`);
- if you don't want the column, drop `TonCorePoolSlotDataSource` from the server DTO — leaving it as a dead wire field is worse than not having it.

I'd take the first.

src/node-control/commands/src/commands/nodectl/config_elections_cmd.rs
```diff
@@ -34,6 +34,12 @@ pub enum ElectionsAction {
     TickInterval(TickIntervalCmd),
     /// Set the max-factor
     MaxFactor(MaxFactorCmd),
+    /// Set the AdaptiveSplit50 staking window — when to start staking and how long to wait for peers.
+    ///
+    /// Only used when the stake policy is `adaptive_split50`; ignored for `minimum`, `split50`,
+    /// and `fixed`.
+    #[command(name = "wait-pct")]
```
Keshoid: **Subcommand name inverted vs. the request.**

Review comment 3228178579 asked for primary = `wait` with `#[command(alias = "wait-pct")]`. What landed is the opposite — primary = `wait-pct`, no `wait` alias — so the shorter ergonomic form isn't reachable and operators have to type the long one.

```rust
#[command(alias = "wait-pct")]
Wait(WaitCmd),
```

Functionally fine, but worth flipping before the release goes out — adding an alias later is a breaking-ish change for muscle memory.

src/node-control/commands/src/commands/nodectl/config_pool_cmd.rs
```diff
@@ -605,23 +737,31 @@ impl PoolDepositValidatorCmd {
...
+        // send stake + fee so the credited validator_amount matches `--amount`.
+        let msg_value_nanotons =
+            deposit_nanotons.saturating_add(pool_messages::DEPOSIT_VALIDATOR_POOL_FEE_NANOTONS);
+
         let gas_reserve: u64 = 2_000_000_000;
-        if wallet_info_data.balance < deposit_nanotons.saturating_add(gas_reserve) {
+        if wallet_info_data.balance < msg_value_nanotons.saturating_add(gas_reserve) {
             anyhow::bail!(
-                "Insufficient wallet balance: {} TON (need {} TON + gas)",
+                "Insufficient wallet balance: {} TON (need {:.9} TON on-chain: {:.9} stake + {:.9} pool fee + gas reserve)",
```
Keshoid: Minor — error message lists `"... {stake} stake + {fee} pool fee + gas reserve"` but `gas reserve` is bare text with no number, so the `X` in `need X TON on-chain` doesn't equal `stake + fee` and the operator has to subtract to learn the gas reserve is 2 TON. Either interpolate the value or drop the term:

```rust
"Insufficient wallet balance: {} TON (need {:.9} TON on-chain: {:.9} stake + {:.9} pool fee + {:.9} TON gas reserve)",
nanotons_to_tons_f64(wallet_info_data.balance),
nanotons_to_tons_f64(msg_value_nanotons.saturating_add(gas_reserve)),
self.amount,
nanotons_to_tons_f64(pool_messages::DEPOSIT_VALIDATOR_POOL_FEE_NANOTONS),
nanotons_to_tons_f64(gas_reserve),
```

src/node-control/commands/src/commands/nodectl/config_pool_cmd.rs
```diff
@@ -273,21 +286,60 @@ struct PoolAddCoreBody<'a> {
     min_nominator_stake: Option<u64>,
 }
 
+fn share_pct_to_bp(pct: f64) -> anyhow::Result<u16> {
+    anyhow::ensure!(
+        pct.is_finite() && (0.0..=100.0).contains(&pct),
```
Keshoid: Minor / domain check — the inclusive upper bound lets `--validator-share-percent 100` through, producing `validator_share = 10000` bp, which on TONCore means the validator takes the entire reward and nominators get 0. The server-side check (`PoolAddCoreBody` handler) accepts `0..=10_000` too, so both sides agree, but it's almost certainly a footgun rather than a legitimate config.

If we want to harden it, the place is here — bound to `< 100.0` (exclusive) and bail with a message pointing operators at the nominator-share implication. Skip if you'd rather leave bp validation to a single domain-aware spot on the server.

src/node-control/commands/src/commands/nodectl/config_pool_cmd.rs
mrnkslv: done

src/node-control/commands/src/commands/nodectl/config_pool_cmd.rs
mrnkslv: done

src/node-control/commands/src/commands/nodectl/config_elections_cmd.rs
mrnkslv: done

src/node-control/commands/src/commands/nodectl/config_pool_cmd.rs
mrnkslv: done

src/node-control/commands/src/commands/nodectl/config_pool_cmd.rs
mrnkslv: done

Keshoid (approved):

Keshoid (commented):
Rust idiom pass — three narrow API/style nits on the new code, all easy fixes.

src/node-control/commands/src/commands/nodectl/config_pool_cmd.rs
```diff
@@ -329,6 +397,76 @@ impl PoolAddCoreCmd {
...
+    fn as_wire(self) -> &'static str {
+        match self {
+            Self::NotDeployed => "not deployed",
+            Self::Active => "active",
+            Self::Frozen => "frozen",
+            Self::Error => "error",
+        }
+    }
+
+    fn from_wire(s: &str) -> Result<Self, ()> {
```
Keshoid: **`Result<Self, ()>` → `Option<Self>`.**

`()` as the error variant is meaningless — there's no information to convey beyond "it failed". The sole caller (`deserialize_toncore_pool_slot_wire_state`) immediately discards it with `.map_err(|_| de::Error::custom(...))`, which is the standard tell that the unit error is dead weight.

The idiomatic shape is `Option<Self>`:

```rust
fn from_wire(s: &str) -> Option<Self> {
    Some(match s {
        "" | "not deployed" => Self::NotDeployed,
        "active" => Self::Active,
        "frozen" => Self::Frozen,
        "error" => Self::Error,
        _ => return None,
    })
}
```

Caller becomes:

```rust
TonCorePoolSlotWireState::from_wire(&s)
    .ok_or_else(|| de::Error::custom(format!("unknown TONCore pool slot state: {s:?}")))
```

Drops one `Result` layer for no information loss.

src/node-control/commands/src/commands/nodectl/config_pool_cmd.rs
```diff
@@ -329,6 +397,76 @@ impl PoolAddCoreCmd {
...
+/// to `"not deployed"` (single multi-word token on the wire).
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum TonCorePoolSlotWireState {
+    NotDeployed,
+    Active,
+    Frozen,
+    Error,
+}
+
+impl Default for TonCorePoolSlotWireState {
```
Keshoid: **The whole `Default` + `Serialize` + custom deserializer block (~40 lines) collapses to a 6-line derive.**

Since Rust 1.62, `#[derive(Default)]` works on enums via `#[default]` on a variant. And serde handles `"not deployed"` (with the space) via `#[serde(rename = ...)]`:

```rust
#[derive(
    Clone, Copy, Debug, Default, PartialEq, Eq,
    serde::Serialize, serde::Deserialize,
)]
#[serde(rename_all = "lowercase")]
enum TonCorePoolSlotWireState {
    #[serde(rename = "not deployed")]
    #[default]
    NotDeployed,
    Active,
    Frozen,
    Error,
}
```

Then on the field:

```rust
#[serde(default)]
state: TonCorePoolSlotWireState,
```

Drops `impl Default`, the custom `Serialize` impl, `as_wire` / `from_wire` helpers, and `deserialize_toncore_pool_slot_wire_state`. The only behavior you lose is the `"" → NotDeployed` fallback, which is defensive coding for a case the server never produces (it always serializes one of the four wire strings). If you want to keep that, `#[serde(alias = "")]` on `NotDeployed` brings it back in one line.

src/node-control/commands/src/commands/nodectl/config_pool_cmd.rs
```diff
@@ -273,21 +286,75 @@ struct PoolAddCoreBody<'a> {
...
+    Ok(bp)
+}
+
+/// Validate `--validator-share` raw basis-point input.
+///
+/// Mirrors the server-side check (`0..=10_000`) but tightens the upper bound
+/// to **`< 10_000`**: a 100 % validator share would leave nominators with no
+/// pool rewards, which is almost always operator error rather than intent.
+fn validate_validator_share_bp(bp: u16) -> anyhow::Result<u16> {
+    anyhow::ensure!(
```
Keshoid: **First `ensure!` is subsumed by the second.**

`bp != 10_000` accepts everything in `0..10_000 ∪ 10_001..=u16::MAX`, then `bp < 10_000` accepts only `0..10_000`. Anything the first rejects (`bp == 10_000`) the second also rejects — so the first guard only exists for a friendlier error message at that specific value, which is fine intent but reads as dead validation on first scan.

A single `match` makes the intent obvious and drops one branch:

```rust
fn validate_validator_share_bp(bp: u16) -> anyhow::Result<u16> {
    match bp {
        10_000 => anyhow::bail!(
            "--validator-share 10000 bp is 100%: nominators would receive no pool rewards — use a value below 10000 bp"
        ),
        bp if bp > 10_000 => anyhow::bail!(
            "validator_share must be in 0..10000 basis points (<100%; got {bp})"
        ),
        bp => Ok(bp),
    }
}
```

src/node-control/commands/src/commands/nodectl/config_pool_cmd.rs
mrnkslv: done

src/node-control/commands/src/commands/nodectl/config_pool_cmd.rs
mrnkslv: done

src/node-control/commands/src/commands/nodectl/config_pool_cmd.rs
mrnkslv: done

PR #132, opened 2026-05-04
Feature/sma 85 migrate vote commands to rest api

mrnkslv:
## Summary

Migrates config-proposal voting to the nodectl control HTTP API: read endpoints for nominators (and operators), operator-only mutations for the tracked proposal list, persistence via update_and_save, and config_changed notifications. The nodectl vote CLI is now a thin REST client (shared URL, token, and config resolution with other REST commands).

## Changes

- HTTP: GET /v1/voting/config, GET /v1/voting/proposals, GET /v1/voting/proposals/{hash}, POST /v1/voting/proposals, DELETE /v1/voting/proposals/{hash} with role-appropriate auth middleware.
- Service: voting DTOs and handlers in config_handlers.rs; route registration and OpenAPI updates in http_server_task.rs; unit tests in config_handlers_tests.rs and auth coverage in auth_tests.rs.
- CLI: vote_cmd.rs uses the service API instead of loading/saving config locally for list, inspect, add, and remove flows.
- Docs: README updates for voting, REST endpoint table, and env flags; CHANGELOG.md entry under Unreleased.
- E2E: run_singlehost_nodectl.py smoke-checks voting REST and nodectl vote ls; optional VOTING_REST_VALIDATE bypass; test_run_net_py/readme.md note.


src/node/tests/test_run_net_py/run_singlehost_nodectl.py
```diff
@@ -417,6 +419,84 @@ def _nctl_output(self, *args: str, check: bool = True, timeout: int = 30) -> str
         )
         return result.stdout
 
+    def _nodectl_rest_base_url(self) -> str:
```
mrnkslv, line 448: replace with nodectl commands

Keshoid (commented):
Solid PR overall — REST handlers mirror the existing CRUD shape, auth split is correct, and the smoke test covers the read paths. A handful of small simplifications and a couple of test gaps below; none blocking.

A single follow-up that addresses items 1, 2, 3, 4, 7, 9 (plus the operator-allowed auth test) should shave ~30–40 lines off the diff with no behavior change.

src/node-control/service/src/http/config_handlers.rs
```diff
@@ -917,14 +993,293 @@ async fn extract_public_key(state: &AppState) -> Option<String> {
...
+// ---------------------------------------------------------------------------
+// Voting API (`/v1/voting/*`)
+//
+// GET routes use `require_nominator`; POST/DELETE use `require_operator`
+// (see `http_server_task::routes`). Mutations use `update_and_save` and
+// `config_changed.notify_one()` like CRUD handlers below.
+// ---------------------------------------------------------------------------
+
+/// Default `tick_interval` when the `voting` section is absent (matches `VotingConfig` serde default).
+const VOTING_TICK_INTERVAL_DEFAULT_SECS: u64 = 40;
```
Keshoid: **Drift risk: duplicate default.** This `40` duplicates the serde default of `VotingConfig::tick_interval`, and the same literal is repeated at line 1216 inside `update_and_save`. If the default in `common::app_config` ever changes, the GET response and newly-created voting sections will silently disagree.

Replace with `VotingConfig::default().tick_interval` at both sites (or expose a `pub const` in `common::app_config` and use it in the serde default). Net: −1 constant, one source of truth.

src/node-control/service/src/http/config_handlers.rs
```diff
@@ -917,14 +993,293 @@ async fn extract_public_key(state: &AppState) -> Option<String> {
...
+
+fn parse_voting_proposal_hash_hex(s: &str) -> Result<[u8; 32], AppError> {
+    let s = s.trim().to_lowercase();
```
Keshoid: `hex::decode` is case-insensitive, so `.to_lowercase()` is an unnecessary allocation. `let s = s.trim();` is enough.

src/node-control/service/src/http/config_handlers.rs
```diff
@@ -917,14 +993,293 @@ async fn extract_public_key(state: &AppState) -> Option<String> {
...
+    let bytes =
+        hex::decode(&s).map_err(|_| AppError::bad_request("proposal hash must be valid hex"))?;
+    if bytes.len() != 32 {
+        return Err(AppError::bad_request(format!(
+            "proposal hash must be 32 bytes (64 hex digits), got {} bytes",
+            bytes.len()
+        )));
+    }
+    let mut out = [0u8; 32];
+    out.copy_from_slice(&bytes);
```
Keshoid: Minor: `let mut out = [0u8; 32]; out.copy_from_slice(&bytes); Ok(out)` can collapse to `Ok(bytes.try_into().expect("len checked above"))` (or fold the length check into the `try_into`). Saves a couple of lines.

src/node-control/service/src/http/config_handlers.rs
```diff
@@ -917,14 +993,293 @@ async fn extract_public_key(state: &AppState) -> Option<String> {
...
+        voters_count: p.voters.len(),
+        weight_remaining: p.weight_remaining,
+        rounds_remaining: p.rounds_remaining,
+        wins: p.wins,
+        losses: p.losses,
+        tracked,
+    }
+}
+
+fn voting_proposal_detail_dto(p: &ConfigProposal) -> VotingProposalDetailDto {
```
Keshoid: Style nit: this takes no extra state and is a straight mapping — make it `impl From<&ConfigProposal> for VotingProposalDetailDto`, which is exactly the pattern that was deleted from `vote_cmd.rs`. The row variant needs `tracked_lower`, so leave it as a free fn (but the `voting_` prefix is redundant inside a voting-only section — `row_dto` reads fine).

src/node-control/service/src/http/config_handlers.rs
```diff
@@ -917,14 +993,293 @@ async fn extract_public_key(state: &AppState) -> Option<String> {
...
+pub async fn v1_voting_proposals_list_handler(
+    state: axum::extract::State<AppState>,
+) -> Result<axum::Json<VotingProposalsListResponse>, AppError> {
+    let cfg = state.runtime_cfg.get();
+    let tracked_lower: Vec<String> = cfg
+        .voting
+        .as_ref()
+        .map(|v| v.proposals.iter().map(|h| h.to_lowercase()).collect())
+        .unwrap_or_default();
+    drop(cfg);
```
Keshoid: Heads-up (not a blocker, this matches the file's existing convention): `RuntimeConfigStore::get()` returns `Arc<AppConfig>` (runtime_config.rs:497–500), not a lock guard, so `drop(cfg);` is a no-op and slightly misleads readers into expecting guard semantics. The same pattern is repeated at lines 1190 and 1259 for the new voting handlers.

We can just remove this `drop()`

src/node-control/service/src/http/config_handlers.rs
```diff
@@ -917,14 +993,293 @@ async fn extract_public_key(state: &AppState) -> Option<String> {
...
+    let hash_hex = hex::encode(hash_bytes);
+
+    let cfg = state.runtime_cfg.get();
+    let in_list = cfg
+        .voting
+        .as_ref()
+        .is_some_and(|v| v.proposals.iter().any(|h| h.eq_ignore_ascii_case(&hash_hex)));
+    if !in_list {
+        return Err(AppError::not_found(format!(
+            "proposal '{hash_hex}' is not in the voting tracked list"
```
Keshoid: Cosmetic: this error quotes the hash (`'{hash_hex}'`), while the add-path error at line 1204 doesn't. Pick one style for consistency.

src/node-control/commands/src/commands/nodectl/vote_cmd.rs
```diff
@@ -359,49 +310,37 @@ impl VoteRmCmd {
...
-    let config = AppConfig::load(Path::new(config_path))?;
-    let rpc_client = Arc::new(
-        ClientJsonRpc::connect_many(
-            config.ton_http_api.resolved_endpoints(),
-            config.ton_http_api.api_key.clone(),
-        )
-        .context("ClientJsonRpc")?,
-    );
-    Ok((config, rpc_client))
+async fn fetch_proposal_rows(
```
Keshoid: Simplification: `fetch_proposal_rows` and `fetch_tracked_hashes` (below) both build a `serde_json::Value`, `.get("result").cloned()`, then `from_value`. A typed envelope removes the clone and the verbose unwraps:

```rust
#[derive(Deserialize)]
struct ApiOk<T> { result: T }

let env: ApiOk<Vec<ProposalRow>> = serde_json::from_str(&body)?;
```

Apply at the four call sites (`VoteLsCmd::run` ~L106, `VoteInspectCmd::run` ~L194, and both `fetch_*` helpers). Removes ~10 lines and an obvious clone. Bonus: `fetch_proposal_rows` is only used by `VoteAddCmd` interactive — once the envelope helper exists, inline it there and drop the helper entirely (~8 more lines).

src/node-control/commands/src/commands/nodectl/vote_cmd.rs
```diff
@@ -359,49 +310,37 @@ impl VoteRmCmd {
...
+) -> anyhow::Result<Vec<ProposalRow>> {
+    let body = api_get(base_url, "/v1/voting/proposals", token).await?;
+    let v: serde_json::Value = serde_json::from_str(&body)?;
+    serde_json::from_value(
+        v.get("result")
+            .cloned()
+            .ok_or_else(|| anyhow::anyhow!("voting proposals: missing 'result'"))?,
+    )
+    .context("parse proposal rows")
 }
```
Keshoid, line 303: `.context("parse proposal rows")` only on this call site — the matching `from_value` in `VoteLsCmd::run` (L106) and `VoteInspectCmd::run` (L194) don't add context. Either drop it here or add equivalents at the other sites.

src/node-control/commands/src/commands/nodectl/vote_cmd.rs
```diff
@@ -359,49 +310,37 @@ impl VoteRmCmd {
...
+        .get("result")
+        .and_then(|r| r.get("proposals"))
+        .and_then(|p| p.as_array())
+        .ok_or_else(|| anyhow::anyhow!("voting config: missing result.proposals"))?;
+    Ok(arr.iter().filter_map(|x| x.as_str().map(str::to_string)).collect())
 }
 
-fn parse_proposal_hash(s: &str) -> anyhow::Result<[u8; 32]> {
-    let bytes = hex::decode(s).context("invalid hex")?;
+fn parse_proposal_hash_hex_normalized(s: &str) -> anyhow::Result<String> {
```
Keshoid: This helper plus the client-side membership pre-check in `VoteRmCmd::run` (L291) duplicate validation the server already performs (400 for invalid hex, 404 for not-tracked). The duplication also creates two error paths that disagree under races.

Simplest fix: drop `parse_proposal_hash_hex_normalized` entirely and pass `self.hash.trim()` (or `self.hash.to_lowercase()` if you want to canonicalise) straight into the URL. Server is authoritative. Removes ~15 lines from this file and one cross-crate duplicate (the equivalent `parse_voting_proposal_hash_hex` already exists in the service).

src/node-control/service/src/http/auth_tests.rs
```diff
@@ -346,6 +346,30 @@ async fn nominator_forbidden_on_operator_route() {
...
+#[tokio::test]
+async fn voting_config_nominator_token_200() {
+    let st = state_with_auth().await;
+    let tok = st.jwt_auth.generate("nom", Role::Nominator, 3600).unwrap().0;
+    let resp = app(st).oneshot(get_bearer("/v1/voting/config", &tok)).await.unwrap();
+    assert_eq!(resp.status(), 200);
+}
+
+#[tokio::test]
+async fn voting_proposals_post_nominator_forbidden() {
```
Keshoid, line 437: Missing the symmetric positive test: there's `voting_proposals_post_nominator_forbidden` here, but no `operator_allowed_on_voting_route` analogue covering POST + DELETE under an operator token. The non-voting equivalent at L374 (`operator_allowed_on_operator_route`) sets the pattern.

Note: the actual POST happy path (chain-existence branch + `update_and_save`) is also untested in `config_handlers_tests.rs` — only the already-tracked short-circuit and the 400 are covered. Consistent with how the rest of the file skips on-chain branches (no `ConfigContractImpl` mock plumbing), so flagging rather than blocking.

Keshoid (changes requested):

src/node/tests/test_run_net_py/run_singlehost_nodectl.py
mrnkslv: done

src/node-control/service/src/http/config_handlers.rs

src/node-control/service/src/http/config_handlers.rs

src/node-control/commands/src/commands/nodectl/vote_cmd.rs
mrnkslv: done

src/node-control/commands/src/commands/nodectl/vote_cmd.rs
mrnkslv: outdated because of previous comment

src/node-control/commands/src/commands/nodectl/vote_cmd.rs
mrnkslv: ready

src/node-control/service/src/http/auth_tests.rs
mrnkslv: added

src/node-control/service/src/http/config_handlers.rs

src/node-control/service/src/http/config_handlers.rs

src/node-control/service/src/http/config_handlers.rs
mrnkslv: ok

src/node-control/service/src/http/config_handlers.rs
mrnkslv: replaces

Keshoid (approved):

PR #138, opened 2026-05-12
feat:Deploy TON Core pool via stub-with-SETCODE init message

mrnkslv:
## Summary
Fix SMA-90 — TON Core nominator pools deployed by nodectl were rejected by Tonscan as `invalid state init of contract`. This PR introduces a per-slot deploy mode that produces a Tonscan-recognized StateInit (bootstrap stub + `SETCODE` on first run, mirroring the canonical `new-pool.fif` flow), while keeping legacy behavior available so already-deployed pool addresses remain stable.

## Changes
- **Common / config (`app_config`):** new `TonCoreDeployMode` enum with variants `Legacy` (full pool code in `StateInit.code`) and `TonscanCompatible` (bootstrap code + `SETCODE`). Per-slot `deploy_mode` field on `TonCorePoolConfig` (JSON key kept as `deploy_layout` for compatibility). `#[non_exhaustive]`, `FromStr` driven by serde, snake_case + alias wire strings.
- **Defaults split** so the bug is fixed for new pools without breaking existing ones:
  - Persisted config missing the field → `Legacy` (stable derived addresses for pools deployed by older nodectl).
  - `POST /v1/pools/core` / `nodectl config pool add core` omitting deploy mode → `TonscanCompatible` (`default_new_pool_deploy_mode`).
- **Contracts (`ton_core_nominator`):** dual `StateInit` paths and layout-aware address derivation. Bootstrap stub is built from named TVM opcode constants (`OP_SETCP0`, `OP_ACCEPT`, `OP_PUSHREF`, `OP_SETCODE`). Unit test asserts stub references the full pool code.
- **Service / REST:** `PoolAddCoreRequest` and `TonCorePoolSlotDto` carry `deploy_layout`; OpenAPI registers `TonCoreDeployMode` so `$ref`s resolve. Handler tests cover (a) default-on-create is `tonscan`, (b) explicit `tonscan` round-trips, (c) persisted-without-field reads as `legacy`.
- **CLI:** `nodectl config pool add core --deploy-mode <legacy|tonscan>` (alias `--deploy-layout`); value parser is `clap::value_parser!(TonCoreDeployMode)` (single source of truth with serde).
- **Docs:** `README.md` and `CHANGELOG.md` document the field, the address-derivation impact, and the recommendation to use `tonscan` for new pools.


src/node-control/service/src/http/config_handlers.rs
```diff
@@ -627,16 +635,16 @@ pub async fn v1_pools_handler(
                 for (idx, slot_cfg) in pools.iter().enumerate() {
                     let Some(cfg) = slot_cfg.as_ref() else { continue };
                     let cached = cached_iter.next();
-                    slot_jobs.push((idx, cfg.address.clone(), cached));
+                    slot_jobs.push((idx, cfg.clone(), cached));
                 }
```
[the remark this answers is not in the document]
mrnkslv: fixed

Keshoid:
**Naming: `TonCoreDeployLayout` → `TonCoreDeployMode` + variant rename**

This field is operator-facing (lives in JSON config and a CLI flag), so the variant names should describe what the operator gets, not how the StateInit is built. `EmbeddedCode` / `ActivateUpgrade` describe the mechanism — an operator scanning the config can't tell which one fixes the tonscan parsing problem this PR is solving.

Proposal:

```rust
#[derive(Clone, Copy, PartialEq, Eq, Debug, Default, serde::Serialize, serde::Deserialize)]
#[cfg_attr(feature = "openapi", derive(utoipa::ToSchema))]
pub enum TonCoreDeployMode {
    #[default]
    #[serde(rename = "legacy")]
    Legacy,
    #[serde(
        rename = "tonscan",
        alias = "tonscan_compatible",
        alias = "tonscan-compatible",
    )]
    TonscanCompatible,
}
```

Why these names:

- `Mode` reads more naturally than `Layout` for a user-facing knob.
- `Legacy` honestly signals "kept only for pools that were already deployed by older nodectl; don't pick this for new deployments" — which is the real semantic, since the default has to stay backwards-compatible with already-derived addresses.
- `TonscanCompatible` tells the operator the one thing they need to know to pick correctly.
- The serde tags (`legacy` / `tonscan`) keep typing short in CLI and JSON; the long form `tonscan_compatible` is still accepted as an alias for readers who copy from the Rust name.

Side benefit: with serde owning the string forms, `parse_cli_deploy_layout` in `config_pool_cmd.rs` can be removed and replaced with `clap::value_parser!(TonCoreDeployMode)` (assuming a `FromStr` impl driven by serde). One source of truth for accepted values, and the help text can list them programmatically.

Help text would then read something like:

```
--deploy-mode <MODE>
    Deploy mode for new TONCore pool slots.
    `tonscan` — deploy via stub + SETCODE; required for tonscan to recognise the pool (recommended for new deployments).
    `legacy`  — full pool code embedded directly in StateInit (kept for compatibility with pools already deployed by older nodectl).
    [default: legacy]
```

If the team decides to flip the default to `tonscan` for new pools, only that line and the `#[default]` attribute change.

Keshoid:
**Default for new TONCore pool creation should be the tonscan-compatible mode**

Right now every entry point defaults to the embedded-code layout — `#[default]` on the enum, `#[serde(default)]` on `PoolAddCoreRequest.deploy_layout`, and `self.deploy_layout.unwrap_or_default()` in the CLI handler. Net effect: a fresh `nodectl config pool add core …` or `POST /v1/pools/core` without an explicit flag produces exactly the StateInit shape tonscan rejects — i.e. the bug SMA-90 is meant to fix is still the out-of-the-box behavior.

The enum's `#[default]` can't simply be flipped, though: it's also what fills the field in when an existing JSON config loaded by an older nodectl version is re-read with the new struct. Those pools were deployed with the legacy layout, and their derived addresses must keep matching, so a missing field has to read as legacy.

Proposal — split "default at deserialization" from "default at creation":

```rust
// Enum: #[default] stays Legacy so old configs keep deriving the same address.
#[derive(Default, …)]
pub enum TonCoreDeployMode {
    #[default]
    Legacy,
    TonscanCompatible,
}

// CLI (config_pool_cmd.rs):
let deploy_mode = self.deploy_mode.unwrap_or(TonCoreDeployMode::TonscanCompatible);

// REST (config_handlers.rs):
fn default_new_pool_deploy_mode() -> TonCoreDeployMode { TonCoreDeployMode::TonscanCompatible }

#[derive(Deserialize, …)]
pub struct PoolAddCoreRequest {
    …
    #[serde(default = "default_new_pool_deploy_mode")]
    pub deploy_mode: TonCoreDeployMode,
}
```

Result:
- Existing JSON configs without the field → reads as `Legacy` → addresses stable for already-deployed pools.
- New pool via CLI or REST without an explicit value → `TonscanCompatible` → tonscan recognises it without the operator needing to know about the knob.

Worth calling out this split explicitly in the CHANGELOG entry too.

Keshoid:
**PR description advertises files that aren't in the diff**

The PR summary's "Scripts / docs" section mentions `testnet_toncore_activate_upgrade_deploy.sh` and `run_testnet_toncore_demo.py`, but neither file appears in the diff. Could you either add them or trim the description? It's confusing for reviewers to figure out which list is authoritative.

src/node-control/contracts/src/nominator/ton_core_nominator_router.rs
```diff
@@ -60,18 +60,24 @@ impl TonCoreNominatorRouter {
         provider: Arc<dyn ContractProvider>,
         pools: [Option<TonCoreInitParams>; 2],
         validator_address: &MsgAddressInt,
+        deploy_layouts: [TonCoreDeployLayout; 2],
```
Keshoid: `TonCoreNominatorRouter::from_state_init` has no callers in the tree — `runtime_config.rs:678` builds the router via `from_wrappers([w0, w1])`. The new `[TonCoreDeployLayout; 2]` parameter is dead code.

Either delete the function (the `from_wrappers` path covers the production use case) or add a test that exercises it so the per-slot layout wiring is verified end-to-end. Leaving it as-is means future refactors have to keep updating an unused signature.

src/node-control/README.md
```diff
@@ -1797,7 +1797,8 @@ Nominator pool configurations. Pool `kind` is **`"snp"`** or **`"core"`**.
...
-  - `address` — optional deployed pool contract address (raw / base64url string). When omitted but `params` is set, the address is derived from the validator and `params` (see `resolve_toncore_pool` / `toncore_pool_address_and_state` in the contracts crate). If `address` is set, it must match the derived address when `params` is present.
+  - `address` — optional deployed pool contract address (raw / base64url string). When omitted but `params` is set, the address is derived from the validator and `params` (see `resolve_toncore_pool` / `toncore_pool_address_and_state` in the contracts crate). If `address` is set, it must match the derived address when `params` is present. Derivation depends on `deploy_layout`: `embedded_code` and `activate_upgrade` yield different addresses for the same parameters.
+  - `deploy_layout` — optional string per slot: `embedded_code` (default — entire pool bytecode in `StateInit.code`) or `activate_upgrade` (minimal bootstrap in `code`, full pool installed via `SETCODE` on first execution — often described as “upgrade on activation”.
```
Keshoid: Small typo — the parenthetical that opens with `(minimal bootstrap in code, …` never closes. Add the trailing `)` after `"upgrade on activation"`.

Also worth adding one sentence here recommending the tonscan-compatible mode for new pools — otherwise an operator reading the config docs has no way to tell which value to pick.

src/node-control/commands/src/commands/nodectl/config_pool_cmd.rs
```diff
@@ -17,14 +17,24 @@ use crate::commands::nodectl::{
...
-    app_config::PoolConfig,
+    app_config::{PoolConfig, TonCoreDeployLayout},
     task_cancellation::CancellationCtx,
     ton_utils::{display_tons, nanotons_to_tons_f64, tons_f64_to_nanotons},
 };
 use contracts::{TonWallet, nominator::ton_core_pool as pool_messages};
 use std::{path::Path, str::FromStr};
 use ton_block::{ADDR_FORMAT_BOUNCE, ADDR_FORMAT_URL_SAFE, MsgAddressInt, write_boc};
 
+fn parse_cli_deploy_layout(s: &str) -> Result<TonCoreDeployLayout, String> {
```
Keshoid: `parse_cli_deploy_layout` hand-rolls a string→enum mapping that serde already owns via `#[serde(rename_all = "snake_case")]`. If anyone renames a variant (e.g. via the naming proposal in the top-level PR comment), the CLI silently accepts stale names without a compile error.

Cheapest fix: drop this function. Either

(a) implement `FromStr` on the enum in `common::app_config` delegating to serde (via `serde_plain::from_str` or a small manual impl next to the enum), and use `value_parser = clap::value_parser!(TonCoreDeployMode)` in the clap derive — single source of truth for accepted strings, and `clap` lists them in `--help` automatically; or

(b) keep a tiny manual function but add a unit test that asserts `parse_cli_…("legacy")` matches `serde_json::from_str::<…>(r#""legacy""#)` for every accepted value, so divergence breaks CI.

(a) is the right one.

src/node-control/CHANGELOG.md
```diff
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- **TONCore pool deploy layout** — per-slot config field `deploy_layout` (`embedded_code` default, or `activate_upgrade` for minimal bootstrap `StateInit.code` with full pool code applied via `SETCODE` on activation). Wired through JSON config, REST (`POST /v1/pools/core`, pool views), and CLI (`config pool add core --deploy-layout …`).
```
Keshoid: The entry describes the new field mechanically but never mentions tonscan — which is the entire reason this feature exists. An operator skimming the changelog has no way to know they should switch modes for new deployments.

Suggestion: add one short sentence explaining the motivation and the recommendation. Something like:

> Use `activate_upgrade` for new pools — tonscan rejects pools deployed with `embedded_code` as `invalid state init of contract`, while `activate_upgrade` matches the canonical `new-pool.fif` flow that tonscan recognises. Pools already deployed with `embedded_code` must keep that value (derivation depends on the layout, so changing it would change the pool address).

If the renaming proposal in the PR comment lands, swap the variant names accordingly.

src/node-control/common/src/app_config.rs
```diff
@@ -341,6 +374,10 @@ pub struct TonCorePoolConfig {
     #[serde(default)]
     #[serde(skip_serializing_if = "Option::is_none")]
     pub params: Option<TonCoreInitParams>,
+    /// Deploy layout for this slot (`embedded_code` when omitted in JSON).
+    #[serde(default)]
+    #[serde(skip_serializing_if = "skip_serializing_toncore_deploy_layout")]
+    pub deploy_layout: TonCoreDeployLayout,
```
Keshoid: `TonCoreInitParams` (`app_config.rs:385`) is a tiny `struct` of two `u16`s and two `u64`s (20 bytes, all fields `Copy`), but it derives only `Clone`. As a result, several call sites currently clone it unnecessarily:

- `utils.rs:90` — `resolve_toncore_pool(..., params.clone(), ...)`
- `ton_core_nominator_router.rs:67` — `let Some(init_params) = pools[idx].clone()` (clones `Option<TonCoreInitParams>`)

Suggestion: add `Copy` to the derive.

```rust
#[derive(serde::Serialize, serde::Deserialize, Clone, Copy, PartialEq, Debug)]
pub struct TonCoreInitParams { … }
```

After that, all the `.clone()` calls above drop out — pass-by-value of a 20-byte `Copy` struct is free. Optionally, the `&TonCoreInitParams` arguments in `toncore_pool_address_and_state` / `resolve_toncore_pool` / `from_init_data` could be unified to take the value directly (currently mixed: `resolve_toncore_pool` takes owned, `toncore_pool_address_and_state` takes a borrow), but that's a separate cleanup.

src/node-control/common/src/app_config.rs
```diff
@@ -332,6 +332,39 @@ fn default_toncore_min_nominator_stake() -> u64 {
...
+
+impl TonCoreDeployLayout {
+    #[inline]
+    pub const fn is_embedded_code(self) -> bool {
+        matches!(self, Self::EmbeddedCode)
+    }
+}
+
+impl Default for TonCorePoolConfig {
+    fn default() -> Self {
```
Keshoid: Manual `impl Default` is redundant here — every field of `TonCorePoolConfig` (`Option<String>`, `Option<TonCoreInitParams>`, `TonCoreDeployLayout`) already implements `Default`. Just derive it:

```rust
#[derive(serde::Serialize, serde::Deserialize, Clone, PartialEq, Debug, Default)]
pub struct TonCorePoolConfig {
    #[serde(default, skip_serializing_if = "Option::is_none")]
    pub address: Option<String>,
    #[serde(default, skip_serializing_if = "Option::is_none")]
    pub params: Option<TonCoreInitParams>,
    #[serde(default, skip_serializing_if = "skip_serializing_toncore_deploy_layout")]
    pub deploy_layout: TonCoreDeployLayout,
}
```

…and delete the `impl Default for TonCorePoolConfig` block. Less code, and the default stays in sync automatically if a new field is added later.

src/node-control/contracts/src/nominator/ton_core_nominator.rs
```diff
@@ -34,19 +35,25 @@ fn toncore_pool_address_from_state_init(state_init: &StateInit) -> anyhow::Resul
...
+}
+
+/// Small bootstrap code cell (`SETCP0` · `ACCEPT` · `PUSHREF` pool bytecode · `SETCODE`).
+fn build_toncore_pool_stub_code(full_pool_code: &Cell) -> anyhow::Result<Cell> {
+    let mut b = BuilderData::new();
+    b.append_raw(&[0xff, 0x00], 16)?; // SETCP0
+    b.append_raw(&[0xf8, 0x00], 16)?; // ACCEPT
+    b.append_raw(&[0x88], 8)?; // PUSHREF
+    b.checked_append_reference(full_pool_code.clone())?;
+    b.append_raw(&[0xfb, 0x04], 16)?; // SETCODE
```
Keshoid: The stub is built from raw byte literals with trailing comments telling the reader which TVM opcode each line is. That works, but it's trust-on-faith: a one-byte typo here is silent — the existing `stub_cell_refs_full_pool_code` test only verifies the cell shape (one reference to the pool code), not the opcode semantics, so a wrong opcode would still pass CI and only surface when an actual deploy fails on-chain.

Suggestion: lift the opcodes to named constants. Reads as documented intent, and they become `grep`-able:

```rust
// TVM opcodes used by the deploy bootstrap stub.
// See https://docs.ton.org/v3/documentation/tvm/instructions
const OP_SETCP0:  [u8; 2] = [0xFF, 0x00];
const OP_ACCEPT:  [u8; 2] = [0xF8, 0x00];
const OP_PUSHREF: u8      = 0x88;
const OP_SETCODE: [u8; 2] = [0xFB, 0x04];

fn build_toncore_pool_stub_code(full_pool_code: &Cell) -> anyhow::Result<Cell> {
    let mut b = BuilderData::new();
    b.append_raw(&OP_SETCP0, 16)?;
    b.append_raw(&OP_ACCEPT, 16)?;
    b.append_raw(&[OP_PUSHREF], 8)?;
    b.checked_append_reference(full_pool_code.clone())?;
    b.append_raw(&OP_SETCODE, 16)?;
    Ok(b.into_cell()?)
}
```

If `ton_block` exposes an opcode enum somewhere, migrating to it is then a one-line swap per constant.

src/node-control/service/src/http/config_handlers.rs
```diff
@@ -678,10 +687,12 @@ pub async fn v1_pools_handler(
 async fn fetch_toncore_slot_dto(
     rpc_client: Arc<ClientJsonRpc>,
     slot_idx: usize,
-    config_address: Option<String>,
+    slot_cfg: TonCorePoolConfig,
     cached: Option<Arc<dyn NominatorWrapper>>,
 ) -> TonCorePoolSlotDto {
     let slot_name = if slot_idx == 0 { "even" } else { "odd" };
+    let deploy_layout = slot_cfg.deploy_layout;
```
Keshoid: `slot_cfg` is owned here and never used again after these two lines, so cloning `Option<String>` is unnecessary. A move via destructuring does the same thing without the allocation copy:

```rust
let TonCorePoolConfig { address: config_address, deploy_layout, .. } = slot_cfg;
let slot_name = if slot_idx == 0 { "even" } else { "odd" };
```

Drops the `.clone()` on `slot_cfg.address` and reads cleaner — the intent is "pull the fields out of the config", which destructuring expresses directly.

src/node-control/common/src/app_config.rs
```diff
@@ -332,6 +332,39 @@ fn default_toncore_min_nominator_stake() -> u64 {
...
+/// How `StateInit.code` is shaped when deploying a TON Core nominator pool.
+///
+/// [`TonCoreDeployLayout::EmbeddedCode`] — full pool bytecode is stored in `StateInit.code` at deploy time.
+/// [`TonCoreDeployLayout::ActivateUpgrade`] — minimal bootstrap in `StateInit.code`; full pool code is applied via `SETCODE` when the contract runs (code upgrade on activation).
+///
+/// Both layouts are typically delivered by an **internal** wallet message carrying `state_init`; they differ only in how `code` is encoded before first execution.
+#[derive(Clone, Copy, PartialEq, Eq, Debug, Default, serde::Serialize, serde::Deserialize)]
+#[cfg_attr(feature = "openapi", derive(utoipa::ToSchema))]
+#[serde(rename_all = "snake_case")]
+pub enum TonCoreDeployLayout {
```
Keshoid: Worth marking this enum `#[non_exhaustive]`:

```rust
#[derive(Clone, Copy, PartialEq, Eq, Debug, Default, serde::Serialize, serde::Deserialize)]
#[cfg_attr(feature = "openapi", derive(utoipa::ToSchema))]
#[serde(rename_all = "snake_case")]
#[non_exhaustive]
pub enum TonCoreDeployMode { … }
```

It's a public enum exposed via `common::app_config`, and there's a realistic chance of a third variant later (another deploy mode if tonscan changes its parser, or a different TVM-side bootstrap strategy). With `#[non_exhaustive]`, downstream `match` statements outside this crate are required to have a catch-all arm, so adding a variant later is not a breaking change.

Inside this crate `match` is still exhaustive, so we don't lose compile-time coverage on the matches we actually rely on (in `runtime_config.rs`, the contract code, etc.).

src/node-control/common/src/app_config.rs
```diff
@@ -332,6 +332,39 @@ fn default_toncore_min_nominator_stake() -> u64 {
     DEFAULT_TONCORE_MIN_NOMINATOR_STAKE
 }
 
+#[inline]
+fn skip_serializing_toncore_deploy_layout(layout: &TonCoreDeployLayout) -> bool {
```
Keshoid: The `#[inline]` here is noise — `skip_serializing_toncore_deploy_layout` is a one-line predicate (`layout.is_embedded_code()`), and `is_embedded_code` itself is already `#[inline] const fn` that just does a `matches!`. LLVM inlines trivially small monomorphic functions like this with no annotation needed, and the extra `#[inline]` only makes sense when crossing crate boundaries for a function that's hot in a benchmark.

Safe to drop the `#[inline]` attribute. (Same for the one on `is_embedded_code` — though that one is at least defensible as it's a public `const fn`.)

src/node-control/commands/src/commands/nodectl/config_pool_cmd.rs
```diff
@@ -17,14 +17,24 @@ use crate::commands::nodectl::{
 };
 use colored::{ColoredString, Colorize};
 use common::{
-    app_config::PoolConfig,
+    app_config::{PoolConfig, TonCoreDeployLayout},
     task_cancellation::CancellationCtx,
```
Keshoid, line 21: Side observation while we're here: this parser calls `.trim()` on input, but serde does not — `serde_json::from_str::<TonCoreDeployLayout>("\" embedded_code \"")` rejects the same string that the CLI accepts. So the JSON config and the CLI flag have subtly different rules about whitespace, with no documented reason.

If this function gets dropped in favour of `clap::value_parser!(TonCoreDeployMode)` (per the other suggestion on this line), the asymmetry resolves itself — clap will use the same serde-derived path. If the function stays, drop the `.trim()` so the two paths agree.

Keshoid (changes requested):

src/node-control/README.md
mrnkslv: done

src/node-control/common/src/app_config.rs

src/node-control/service/src/http/config_handlers.rs

src/node-control/common/src/app_config.rs

src/node-control/common/src/app_config.rs

src/node-control/commands/src/commands/nodectl/config_pool_cmd.rs
mrnkslv: done

src/node-control/contracts/src/nominator/ton_core_nominator.rs
mrnkslv: done

src/node-control/common/src/app_config.rs

src/node-control/CHANGELOG.md
mrnkslv: done

src/node-control/commands/src/commands/nodectl/config_pool_cmd.rs
mrnkslv: done

src/node-control/contracts/src/nominator/ton_core_nominator_router.rs
mrnkslv: done

Keshoid (approved):

PR #139, opened 2026-05-14
feat:process TONCore pool withdraw requests

mrnkslv:
## Summary
The elections runner now drains pending nominator withdraw requests from the active TONCore pool before submitting a new stake. This prevents liquidity that nominators have already requested for withdrawal from being re-staked, and reuses the existing process_withdraw_requests opcode (op = 2) on the TONCore nominator-pool contract.

### crate contracts

PoolData extended with withdraw_requests: Option<Cell>; the TONCore get_pool_data wrapper now reads index 10 of the getter stack via cell_opt.
NominatorWrapper trait extended with two methods, both with safe defaults for non-TONCore pools:
has_pending_withdraws() -> Result<bool> (default Ok(false)).
send_process_withdrawals(wallet, query_id, limit, gas_value) -> Result<Cell> (default bail!).
TonCoreNominatorWrapper implements both directly; TonCoreNominatorRouter delegates to its active_pool() so probe and send always target the same slot.
SingleNominatorWrapper keeps withdraw_requests: None (SNP has no withdraw queue, so we don't touch the SNP getter layout).
### crate service — elections runner

Node gets withdraw_processed_for_cycle: Option<u64>; compared by equality with the current election_id, so a new cycle automatically re-arms the step. Cleared in reset_participation (shutdown).
New ElectionRunner::process_pending_withdraws_if_needed, called between recover_stake and participate for every node:
early-returns for non-TONCore pools and when the flag already matches the current cycle;
non-fatal has_pending_withdraws RPC errors are logged and skipped, never blocking participation;
on success the flag is set and the runner continues — participate is intentionally skipped this tick to give the pool time to drain its queue;
on send failure the flag is not set, so the step is retried next tick; the error is surfaced via last_error but does not block stake submission this tick.
New constants: WITHDRAW_PROCESS_GAS = 0.5 TON (mirrors contracts_task::POOL_OP_GAS) and WITHDRAW_PROCESS_LIMIT: u8 = 100 (safe ceiling above pool's max_nominators ≤ 40).
### crate common — snapshot

New ParticipationStatus::ProcessingWithdraws variant (serde processing_withdraws), surfaced for the gap between the opcode-2 submission and the stake message that follows next tick. Anchored to the current election_id so a stale flag from a previous cycle never appears in the snapshot.
### Tests

mock! SingleNominatorWrapper extended with the two new trait methods.
setup_toncore_nominator_slot defaults has_pending_withdraws to Ok(false) so existing TONCore tests are unaffected; setup_toncore_nominator_slot_no_withdraw_defaults lets the new tests wire those expectations explicitly.
Three new tests cover the branch:
test_toncore_pending_withdraws_sends_opcode_and_skips_stake_this_tick — happy path: opcode-2 sent, flag set, stake skipped this tick.
test_toncore_pending_withdraws_skipped_when_already_processed_for_cycle — flag set, probe is not re-issued, stake is submitted normally.
test_toncore_pending_withdraws_send_failure_keeps_flag_unset_and_proceeds — send failure does not set the flag, error surfaces in last_error, stake is still submitted.
### Docs

- `README`: new TONCore-nominator note in the elections-runner section explaining the step and the processing_withdraws participant status.
- `CHANGELOG.md`: new ## [0.5.0] section with an Added entry summarising the feature.

Closes SMA-92


src/node-control/service/src/elections/runner.rs
```diff
@@ -929,6 +964,78 @@ impl ElectionRunner {
...
+                return Ok(false);
+            }
+        };
+        if !has_pending {
+            return Ok(false);
+        }
+
+        let fee = WITHDRAW_PROCESS_GAS + WALLET_COMPUTE_FEE;
+        let wallet_balance = node.wallet_balance().await?;
+        if wallet_balance < fee {
```
Keshoid, line 1009: `bail!` here is the wrong control flow: the caller catches the `Err`, sets `last_error`, and still falls through to `participate`, which will hit the same low-balance condition and log a second error. The opcode-2 step is supposed to be best-effort/skippable, not a hard failure that the caller has to recover from.

Replace with `tracing::warn!` + `return Ok(false)` so we cleanly skip the step and let `participate` decide on its own merits. (`recover_stake` uses `bail!` for the same condition, but there the caller `continue`s the whole node — that pattern doesn't apply here.)

src/node-control/service/src/elections/runner.rs
```diff
@@ -929,6 +964,78 @@ impl ElectionRunner {
...
+        election_id: u64,
+    ) -> anyhow::Result<bool> {
+        let node = self.nodes.get_mut(node_id).expect("node not found");
+        let Some(pool) = node.pool.clone() else {
+            return Ok(false);
+        };
+        if pool.pool_kind() != PoolKind::TONCore {
+            return Ok(false);
+        }
+        if node.withdraw_processed_for_cycle == Some(election_id) {
```
Keshoid: The flag stops us from re-probing within the same cycle, but a nominator can submit a fresh withdraw request right after we send op = 2. Then `has_withdraw_requests` is true again, the pool will reject our stake submission (it doesn't allow new stakes while `withdraw_requests` is non-empty), and our flag prevents us from sending op = 2 again. Net result: we silently miss the cycle.

Proposal: drop `withdraw_processed_for_cycle` entirely. Probe `has_withdraw_requests` every tick where stake is not yet submitted. If pending, send op = 2 and `continue` (skip stake this tick — same as today). Next tick, probe again. Repeat until queue is empty, then stake. Snapshot status follows from the probe directly. `has_withdraw_requests` becomes the single source of truth.

src/node-control/contracts/src/nominator/ton_core_nominator.rs
```diff
@@ -238,8 +243,24 @@ impl NominatorWrapper for TonCoreNominatorWrapper {
             validator_set_changes_count,
             validator_set_change_time,
             stake_held_for,
+            withdraw_requests,
         })
     }
+
+    async fn has_pending_withdraws(&self) -> anyhow::Result<bool> {
+        Ok(self.get_pool_data().await?.withdraw_requests.is_some())
```
Keshoid: Don't pull the entire pool data over RPC just to check whether the dict is non-empty. Add a dedicated `has_withdraw_requests()` getter to the TONCore pool wrapper that calls the pool's cheap on-chain getter and returns `bool` directly. SNP keeps the trait default (`Ok(false)`); TONCore uses the cheap getter.

With the proposed change in the runner (drop the cycle flag, probe every tick), the per-tick cost of this matters: N nodes × every tick × full `get_pool_data` is a lot of wasted bandwidth for a one-bit answer.

src/node-control/contracts/src/nominator/wrapper.rs
```diff
@@ -54,6 +54,27 @@ pub trait NominatorWrapper: SmartContract + Send + Sync {
     fn storage_reserve(&self) -> u64;
     /// Pool type for routing/optimization decisions.
     fn pool_kind(&self) -> PoolKind;
+    /// Whether the pool currently has at least one pending nominator withdraw request.
+    ///
+    /// TONCore-only signal (the pool's `withdraw_requests` dict in persistent storage). For
+    /// pools without a withdraw queue (SNP, direct staking) the default returns `Ok(false)`.
+    async fn has_pending_withdraws(&self) -> anyhow::Result<bool> {
```
Keshoid: Per Rust API guidelines, fallible public trait methods should include an `# Errors` section in their doc comment. Both `has_pending_withdraws` and `send_process_withdrawals` are missing it.

Keshoid (commented):
Inline review notes.

src/node-control/contracts/src/nominator/wrapper.rs
```diff
@@ -54,6 +54,27 @@ pub trait NominatorWrapper: SmartContract + Send + Sync {
...
+        Ok(false)
+    }
+    /// Build an external message that carries `process_withdraw_requests` (TONCore op = 2)
+    /// to this pool. Caller is responsible for ensuring it is meaningful (e.g. checked
+    /// [`has_pending_withdraws`](Self::has_pending_withdraws) and is in the window between
+    /// stake recovery and the next stake submission). Default impl errors out for pool
+    /// kinds that do not support a withdraw queue.
+    async fn send_process_withdrawals(
```
Keshoid: Naming is drifting across three forms: `withdraws` (this trait), `withdrawals` (`send_process_withdrawals`), and `withdraw_requests` (the on-chain op and storage field).

Recommend aligning everything with the on-chain identifier `withdraw_requests`, since that's the canonical name in `pool.fc`:

- `has_pending_withdraws` → `has_withdraw_requests`
- `send_process_withdrawals` → `send_process_withdraw_requests`
- `withdraw_processed_for_cycle` → (drops with the runner-level proposal)
- `ProcessingWithdraws` → `ProcessingWithdrawRequests`

Pick one form — `withdraw_requests` matches the contract exactly, so future readers can grep across both layers.

src/node-control/service/src/elections/runner_tests.rs
```diff
@@ -569,6 +577,23 @@ fn setup_toncore_nominator_slot(
...
+    pool.expect_pool_kind().returning(|| PoolKind::TONCore);
+    // The runner probes pending withdraws between `recover_stake` and `participate`.
+    // Default to "no pending" so existing TONCore tests are unaffected; specific tests
+    // for the new branch use [`setup_toncore_nominator_slot_no_withdraw_defaults`] instead.
+    pool.expect_has_pending_withdraws().returning(|| Ok(false));
+}
+
+/// Same as [`setup_toncore_nominator_slot`] but lets the test wire `has_pending_withdraws`
+/// and `send_process_withdrawals` explicitly. Used by the process-withdraws branch tests.
+fn setup_toncore_nominator_slot_no_withdraw_defaults(
```
Keshoid: `setup_toncore_nominator_slot_no_withdraw_defaults` is a copy of `setup_toncore_nominator_slot` minus the `expect_has_pending_withdraws` line. Replace with a single helper:

```rust
fn setup_toncore_nominator_slot_with(
    pool: &mut MockSingleNominatorWrapper,
    addr: MsgAddressInt,
    state: i32,
    default_has_pending: Option<bool>,
) { ... }
```

and have the existing helper call it with `Some(false)`. Tests for the new branch pass `None` and wire `expect_has_pending_withdraws` themselves.

src/node-control/service/src/elections/runner.rs
```diff
@@ -542,6 +556,27 @@ impl ElectionRunner {
...
+                        node_id
+                    );
+                    continue;
+                }
+                Ok(false) => {}
+                Err(e) => {
+                    if let Some(node) = self.nodes.get_mut(&node_id) {
+                        node.last_error = Some(format!("{:#}", e));
+                    }
+                    tracing::error!("node [{}] process_withdraw_requests error: {:#}", node_id, e);
```
Keshoid: Two nits on this line:

1. `tracing::error!` → `tracing::warn!`. The match arm is explicitly non-fatal (we proceed to `participate`), so `error!` overstates the severity and clutters dashboards.
2. Drop the function name from the message. `process_withdraw_requests error: ...` mixes Rust function naming into operator-facing logs. Use a short component header instead:

```rust
tracing::warn!("node [{}] withdraw requests error: {:#}", node_id, e);
```

Function names rot when refactoring; component-level labels don't.

src/node-control/service/src/elections/runner.rs
```diff
@@ -929,6 +964,78 @@ impl ElectionRunner {
...
     }
 
+    /// Send `process_withdraw_requests` (TONCore op = 2) once per election cycle when the pool
+    /// has at least one pending nominator withdraw. Returns `Ok(true)` only when the message
+    /// was actually broadcast (caller should skip `participate` for this tick to give the pool
+    /// time to drain its queue before computing the new stake amount).
+    ///
+    /// Non-fatal cases — pool kind is not TONCore, queue is empty, opcode already sent for
+    /// this cycle, or `has_pending_withdraws` RPC failed transiently — return `Ok(false)`.
+    async fn process_pending_withdraws_if_needed(
```
Keshoid: Rename: `process_pending_withdraws_if_needed` → `process_pending_withdraw_requests`. Drops the noise suffix and aligns with the contract op name (`process_withdraw_requests`). See the broader naming comment on `wrapper.rs`.

src/node-control/service/src/elections/runner.rs
```diff
@@ -929,6 +964,78 @@ impl ElectionRunner {
...
+        &mut self,
+        node_id: &str,
+        election_id: u64,
+    ) -> anyhow::Result<bool> {
+        let node = self.nodes.get_mut(node_id).expect("node not found");
+        let Some(pool) = node.pool.clone() else {
+            return Ok(false);
+        };
+        if pool.pool_kind() != PoolKind::TONCore {
```
Keshoid: This `pool_kind` guard is redundant. The trait default for `has_pending_withdraws` already returns `Ok(false)` for SNP (and any non-TONCore pool), so the next early-return after the probe handles SNP correctly with zero RPC cost. Drop these three lines and let the trait dispatch.

Keshoid (changes requested):

src/node-control/contracts/src/nominator/wrapper.rs
mrnkslv: added

src/node-control/service/src/elections/runner.rs
mrnkslv: done

src/node-control/service/src/elections/runner_tests.rs
mrnkslv: done

src/node-control/service/src/elections/runner.rs
mrnkslv: done

src/node-control/service/src/elections/runner.rs
mrnkslv: done

src/node-control/contracts/src/nominator/wrapper.rs
mrnkslv: changed

src/node-control/service/src/elections/runner.rs
mrnkslv: done

src/node-control/service/src/elections/runner.rs
mrnkslv: done

src/node-control/contracts/src/nominator/ton_core_nominator.rs
mrnkslv: done

Keshoid (approved):

PR #142, opened 2026-05-15
feat(secrets-vault): add mTLS support for HashiCorp Vault

rayanebel:
The secrets-vault library currently creates a plain `reqwest::Client::new()` when connecting to HashiCorp Vault. This fails when Vault is behind an ingress that requires mutual TLS (mTLS) client certificate authentication.

This PR adds support for the standard HashiCorp Vault TLS environment variables:

  - `VAULT_CACERT` — path to a custom CA certificate (PEM)
  - `VAULT_CLIENT_CERT` — path to a client certificate (PEM)
  - `VAULT_CLIENT_KEY` — path to a client private key (PEM)

When these variables are set, the HTTP client is built with the corresponding TLS configuration. When they are not set, the client behaves exactly as before (no breaking change).

### DIFF
--- a/src/secrets-vault/Cargo.toml
+++ b/src/secrets-vault/Cargo.toml
@@ -61,7 +61,7 @@ tokio = { version = "1", features = [
 ] }
 uuid = { version = "1", features = ["v4", "serde"] }
 zeroize = "1.8"
-reqwest = { version = "0.12", features = ["json"] }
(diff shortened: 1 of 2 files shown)


ITBear (approved):

ITBear:
New PR with fixes: https://github.com/RSquad/ton-rust-node/pull/143

PR #176, opened 2026-05-22
feat(nodectl): refresh past_elections cache periodically

Keshoid:
## Summary

- Adds a TTL-based refresh for the elections runner's `past_elections` and pool-address caches (currently invalidated only on `election_id` change).
- Defends against the SMA-98 incident where a stale snapshot from a lagging RPC endpoint persisted for the whole round and inflated `frozen_stake` to 2× the real value.
- New config field `elections.cache_refresh_secs` (default `300`, set `0` to disable). Hot-reloaded via the 10s config-file watch.
- `Clock` trait introduced in `common` (production `SystemClock`, deterministic `MockClock`) so the TTL logic is unit-testable without real-time sleeps.
- Singlehost e2e (`run_singlehost_nodectl.py`) sets `cache_refresh_secs=60` after `config generate` so the periodic refresh is exercised in the e2e log.

Closes SMA-98.

### DIFF
--- a/src/node-control/service/src/elections/runner_tests.rs
+++ b/src/node-control/service/src/elections/runner_tests.rs
@@ -9,6 +9,7 @@
 use super::*;
 use common::{
     app_config::{ElectionsConfig, NodeBinding, StakePolicy},
+    clock::MockClock,
     snapshot::SnapshotStore,
     task_cancellation::{CancellationCtx, CancellationReason},
     time_format,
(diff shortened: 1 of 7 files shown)


ITBear (approved):

src/node-control/service/src/elections/runner_tests.rs
```diff
@@ -3534,3 +3538,183 @@ async fn test_static_adnl_generation_failure_aborts_node_only() {
     let saved = persisted.lock().unwrap();
     assert!(saved.is_empty(), "nothing should be persisted on failure");
 }
+
+// =====================================================
+// TTL-based cache refresh (SMA-98)
```
mrnkslv, line 3543: i'm not sure if we need that comment

mrnkslv (approved):

PR #185, opened 2026-05-26
fix(node-control): skip redundant TONCore validator set updates

mrnkslv:
## Summary

Compare the TONCore pool saved validator set hash with config param 34 before sending update_validator_set (op 6).

This avoids repeated no-op masterchain transactions on every automation tick and only sends op 6 when the validator set hash actually changed.

### DIFF
--- a/src/node-control/service/src/contracts/contracts_task.rs
+++ b/src/node-control/service/src/contracts/contracts_task.rs
@@ -600,6 +600,8 @@ impl ContractsMonitor {
                                 pool_data.validator_set_changes_count,
                                 now.saturating_sub(last_sent_at),
                             );
+                            // Pool still needs update_validator_set, but fallback is rate-limited.
+                            all_updated = false;
                             continue;
                         }
                     }
@@ -1176,6 +1178,33 @@ mod tests {
         let _ = monitor.ensure_pool_validator_sets_updated().await.unwrap();
         let third_calls = server.send_boc_calls.load(Ordering::Relaxed);
         assert_eq!(third_calls, 2);
+
+        server.shutdown().await;
+    }
+
+    #[tokio::test]
+    async fn ensure_pool_validator_sets_updated_marks_not_ready_when_fallback_is_rate_limited() {
+        let server =
+            MockRpcServer::start_with_config_param_mode("active", 0, Cell::default(), true).await;
+
+        let wallet: Arc<dyn TonWallet> =
+            Arc::new(DummyWallet { addr: addr(3), state_init: Some(StateInit::default()) });
+        let wallets = Arc::new(HashMap::from([("node-a".to_string(), wallet.clone())]));
+        let pools =
+            Arc::new(HashMap::from([("node-a".to_string(), toncore_pool_binding([0u8; 32]))]));
+        let master_wallet: Arc<dyn TonWallet> =
+            Arc::new(DummyWallet { addr: addr(9), state_init: Some(StateInit::default()) });
+
+        let mut monitor =
+            build_monitor_with_pools(server.url.clone(), master_wallet, wallets, pools);
+
+        // First call sends fallback op6 and records the timestamp.
+        let _ = monitor.ensure_pool_validator_sets_updated().await.unwrap();
+
+        // Second call is rate-limited fallback: no send, but still should report not-ready.
+        let all_updated = monitor.ensure_pool_validator_sets_updated().await.unwrap();
+        assert!(!all_updated);
+        assert_eq!(server.send_boc_calls.load(Ordering::Relaxed), 1);
 
         server.shutdown().await;
     }
--- a/src/node-control/service/src/contracts/contracts_task.rs
+++ b/src/node-control/service/src/contracts/contracts_task.rs
@@ -1107,6 +1107,14 @@ mod tests {
 
         assert!(!all_updated);
         assert_eq!(server.send_boc_calls.load(Ordering::Relaxed), 1);
+        assert!(
+            monitor
+                .last_update_sent_at
+                .get("node-a")
+                .and_then(|by_pool| by_pool.get(&addr(4)))
+                .is_some(),
+            "send on Available branch must record last_update_sent_at"
+        );
 
         server.shutdown().await;
     }
--- a/src/node-control/service/src/contracts/contracts_task.rs
+++ b/src/node-control/service/src/contracts/contracts_task.rs
@@ -11,12 +11,12 @@ use anyhow::Context;
 use common::{
     app_config::{AppConfig, ContractsAutomationConfig},
     task_cancellation::CancellationCtx,
+    time_format,
 };
 use contracts::{
     NominatorWrapper, PoolKind, TonWallet, contract_provider,
     nominator::ton_core_pool as tc_messages,
 };
-use control_client::config_params::config_param_cell_repr_hash;
 use std::{
     collections::{HashMap, HashSet},
     sync::Arc,
@@ -32,9 +32,18 @@ const WALLET_GAS: u64 = 100_000_000; // 0.1 TON
 /// Masterchain gas prices are ~25x basechain; 0.1 TON is not enough
 /// for load_data + get_current_validator_set + cell_hash + save_data.
 const POOL_OP_GAS: u64 = 500_000_000; // 0.5 TON
+/// Fallback op6 rate limit when validator-set hash is unavailable.
+const UPDATE_VALIDATOR_SET_FALLBACK_INTERVAL_SEC: u64 = 10 * 60;
 
-fn should_send_update_validator_set(saved_hash: &[u8; 32], current_hash: &[u8; 32]) -> bool {
-    saved_hash != current_hash
+#[derive(Clone, Copy)]
+enum VsetHashState {
+    Unknown,
+    Available([u8; 32]),
+    Unavailable,
+}
+
+fn fallback_update_key(node_id: &str, pool_addr: &MsgAddressInt) -> String {
+    format!("{node_id}:{pool_addr}")
 }
 
 pub(crate) async fn run(
@@ -46,7 +55,14 @@ pub(crate) async fn run(
     let pools = runtime_cfg.pools();
     let wallets = runtime_cfg.wallets();
     let rpc_client = runtime_cfg.rpc_client();
-    let monitor = ContractsMonitor { master_wallet, pools, wallets, rpc_client, runtime_cfg };
+    let mut monitor = ContractsMonitor {
+        master_wallet,
+        pools,
+        wallets,
+        rpc_client,
+        runtime_cfg,
+        fallback_update_sent_at: HashMap::new(),
+    };
     monitor.run_loop(cancellation_ctx).await
 }
 
@@ -56,6 +72,7 @@ struct ContractsMonitor {
     wallets: Arc<HashMap<String, Arc<dyn TonWallet>>>,
     rpc_client: Arc<ClientJsonRpc>,
     runtime_cfg: Arc<dyn RuntimeConfig>,
+    fallback_update_sent_at: HashMap<String, u64>,
 }
 
 impl ContractsMonitor {
@@ -63,7 +80,7 @@ impl ContractsMonitor {
         self.runtime_cfg.get().automation.clone()
     }
 
-    async fn run_loop(&self, cancellation_ctx: CancellationCtx) -> anyhow::Result<()> {
+    async fn run_loop(&mut self, cancellation_ctx: CancellationCtx) -> anyhow::Result<()> {
         let mut cancel = cancellation_ctx.subscribe();
         let mut tick_secs = self.automation().tick_interval_sec.max(1);
         let mut ticker = time::interval(Duration::from_secs(tick_secs));
@@ -95,7 +112,7 @@ impl ContractsMonitor {
         }
     }
 
-    async fn run(&self) -> anyhow::Result<()> {
+    async fn run(&mut self) -> anyhow::Result<()> {
         let auto = self.automation();
         // Master is only needed for deploy/top-up paths. TonCore `update_validator_set` uses the
         // per-node validator wallet, so observe-only mode can run without a funded master.
@@ -485,10 +502,11 @@ impl ContractsMonitor {
     ///
     /// Skips op 6 when config param 34 cell hash matches the pool's `saved_validator_set_hash`
     /// (same check as on-chain op 6).
-    async fn ensure_pool_validator_sets_updated(&self) -> anyhow::Result<bool> {
+    async fn ensure_pool_validator_sets_updated(&mut self) -> anyhow::Result<bool> {
         let provider = contract_provider!(self.rpc_client.clone());
         let mut all_updated = true;
-        let mut current_vset_hash = None;
+        let mut vset_hash_state = VsetHashState::Unknown;
+        let now = time_format::now();
         for (node_id, pool_binding) in
             self.pools.iter().filter(|(_, b)| b.pool_kind() == PoolKind::TONCore)
         {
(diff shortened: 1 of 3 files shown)


src/node-control/service/src/contracts/contracts_task.rs
```diff
@@ -523,6 +532,39 @@ impl ContractsMonitor {
...
+                ) {
+                    tracing::debug!(
+                        target: "contracts",
+                        "[{}] skip update_validator_set: pool={}, vsc_count={}, validator set unchanged",
+                        node_id,
+                        pool_addr,
+                        pool_data.validator_set_changes_count,
+                    );
+                    continue;
+                }
```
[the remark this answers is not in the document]
mrnkslv: added

src/node-control/ton-http-api-client/src/v2/client_json_rpc.rs
```diff
@@ -365,10 +365,13 @@ impl ClientJsonRpc {
             .ok_or_else(|| anyhow::anyhow!(r#"missing "config.bytes" string"#))?;
 
         let boc = base64::engine::general_purpose::STANDARD.decode(b64)?;
-        let cell = read_boc(boc)?.withdraw_single_root()?;
+        read_boc(boc)?.withdraw_single_root().context("config param BOC root")
```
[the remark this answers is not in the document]
mrnkslv: accepted

src/node-control/service/src/contracts/contracts_task.rs
```diff
@@ -523,6 +532,39 @@ impl ContractsMonitor {
...
                 }
 
+                let current_vset_hash = match current_vset_hash {
+                    Some(hash) => hash,
+                    None => match self.rpc_client.get_config_param_cell(34).await {
+                        Ok(cell) => {
+                            let hash = config_param_cell_repr_hash(&cell);
+                            current_vset_hash = Some(hash);
+                            hash
+                        }
```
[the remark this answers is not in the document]
mrnkslv: fixed

src/node-control/service/src/contracts/contracts_task.rs
```diff
@@ -523,6 +532,39 @@ impl ContractsMonitor {
...
+                    &current_vset_hash,
+                ) {
+                    tracing::debug!(
+                        target: "contracts",
+                        "[{}] skip update_validator_set: pool={}, vsc_count={}, validator set unchanged",
+                        node_id,
+                        pool_addr,
+                        pool_data.validator_set_changes_count,
+                    );
+                    continue;
```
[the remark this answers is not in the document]
mrnkslv: done

src/node-control/service/src/contracts/contracts_task.rs
```diff
@@ -477,9 +499,14 @@ impl ContractsMonitor {
     ///
     /// The message is sent from the **validator wallet** for the pool's node (`self.wallets[node_id]`),
     /// not the master wallet — by staking time that wallet is deployed and typically topped up.
-    async fn ensure_pool_validator_sets_updated(&self) -> anyhow::Result<bool> {
+    ///
+    /// Skips op 6 when config param 34 cell hash matches the pool's `saved_validator_set_hash`
+    /// (same check as on-chain op 6).
```
Keshoid: The doc comment describes the happy-path skip but not the fallback. Worth one extra line, e.g.: "When config param 34 cannot be fetched, falls back to a per-pool rate-limited send every `UPDATE_VALIDATOR_SET_FALLBACK_INTERVAL_SEC`." A future reader shouldn't have to derive the intent from the body.

src/node-control/service/src/contracts/contracts_task.rs
```diff
@@ -477,9 +499,14 @@ impl ContractsMonitor {
...
+    async fn ensure_pool_validator_sets_updated(&mut self) -> anyhow::Result<bool> {
         let provider = contract_provider!(self.rpc_client.clone());
         let mut all_updated = true;
+        let mut vset_hash_state = VsetHashState::Unknown;
+        let now = time_format::now();
```
Keshoid: `now` is captured at function entry but used much later (after seqno fetch, get_pool_data, get_config_param, broadcast). On a long tick across many pools, `now` is stale by the time `fallback_update_sent_at.insert(..., now)` runs, biasing rate-limit comparisons earlier on every subsequent tick. Cheap fix: take `time_format::now()` immediately before the insert (after `broadcast` succeeds).

src/node-control/service/src/contracts/contracts_task.rs
```diff
@@ -501,6 +528,7 @@ impl ContractsMonitor {
 
             for pool in pool_binding.inner_pools() {
                 let pool_addr = pool.address().await?;
+                let fallback_key = fallback_update_key(node_id, &pool_addr);
```
Keshoid: `fallback_update_key` allocates a fresh `String` per pool per tick (and `HashMap::get(&String)` requires the same allocation). Consider `(String, MsgAddressInt)` or `Arc<str>`-keyed tuple to avoid the `format!` in the hot path. Not critical at tick intervals measured in seconds, but trivially better and avoids `anti-format-hot-path` patterns.

src/node-control/service/src/contracts/contracts_task.rs
```diff
@@ -31,6 +32,19 @@ const WALLET_GAS: u64 = 100_000_000; // 0.1 TON
 /// Masterchain gas prices are ~25x basechain; 0.1 TON is not enough
 /// for load_data + get_current_validator_set + cell_hash + save_data.
 const POOL_OP_GAS: u64 = 500_000_000; // 0.5 TON
+/// Fallback op6 rate limit when validator-set hash is unavailable.
+const UPDATE_VALIDATOR_SET_FALLBACK_INTERVAL_SEC: u64 = 10 * 60;
+
+#[derive(Clone, Copy)]
```
Keshoid: Minor: `#[derive(Clone, Copy)]` on a 33-byte enum (1-byte discriminant + `[u8; 32]`) is fine today, but if the payload ever grows or gains non-Copy fields, prefer passing by reference. Not actionable now.

src/node-control/service/src/contracts/contracts_task.rs
```diff
@@ -51,14 +72,15 @@ struct ContractsMonitor {
     wallets: Arc<HashMap<String, Arc<dyn TonWallet>>>,
     rpc_client: Arc<ClientJsonRpc>,
     runtime_cfg: Arc<dyn RuntimeConfig>,
+    fallback_update_sent_at: HashMap<String, u64>,
```
Keshoid: Minor: `fallback_update_sent_at` never shrinks. Bounded by pool count, so a non-issue in practice. Just flagging in case pools ever get hot-removed at runtime — at that point a periodic prune (or reconciling against `self.pools` keys) would be worth adding.

Keshoid (commented):
Code review focused on style/perf nits and a small staleness bug. Logic and tests look good; items below are non-blocking except possibly (3).

src/node-control/service/src/contracts/contracts_task.rs
```diff
@@ -31,6 +32,19 @@ const WALLET_GAS: u64 = 100_000_000; // 0.1 TON
 /// Masterchain gas prices are ~25x basechain; 0.1 TON is not enough
 /// for load_data + get_current_validator_set + cell_hash + save_data.
 const POOL_OP_GAS: u64 = 500_000_000; // 0.5 TON
+/// Fallback op6 rate limit when validator-set hash is unavailable.
+const UPDATE_VALIDATOR_SET_FALLBACK_INTERVAL_SEC: u64 = 10 * 60;
```
Keshoid, line 37: Minor: `10 * 60` is a reasonable default for a recovery path, but hardcoding it means test/staging environments can't tune it without a rebuild. I'd leave it as-is for now (avoid speculative config) and only surface it via `ContractsAutomationConfig` when someone actually needs it.

src/node-control/service/src/contracts/contracts_task.rs
```diff
@@ -523,6 +551,62 @@ impl ContractsMonitor {
...
+                            );
+                            vset_hash_state = VsetHashState::Unavailable;
+                            None
+                        }
+                    },
+                };
+
+                if let Some(current_hash) = current_hash {
+                    if current_hash == pool_data.saved_validator_set_hash {
+                        tracing::debug!(
```
Keshoid, line 568: Minor: `tracing::debug!` is correct here — this will be the steady-state path. The two `warn!`/`info!` calls around the fallback are well-placed. No change needed.

src/node-control/service/src/contracts/contracts_task.rs
```diff
@@ -523,6 +551,62 @@ impl ContractsMonitor {
                     continue;
                 }
 
+                let current_hash = match vset_hash_state {
```
Keshoid: Suggestion: hoist the `get_config_param_cell(34)` fetch above the outer pool loop, drop `VsetHashState`, and record the send timestamp unconditionally. Three benefits:

1. The three-state enum is just memoization within a single tick. Fetching once eagerly collapses it to `Option<[u8; 32]>` and flattens the control flow.
2. Recording `fallback_update_sent_at` on **every** successful broadcast (not only on the Unavailable branch) prevents duplicate op6 sends when a previous op6 is still in flight: in tick T we send because hashes differ, in tick T+Δ the contract hasn't processed it yet so `saved_hash` is still stale — without a timestamp we'd send again. With unified bookkeeping the next tick correctly waits.
3. The decision block collapses into `if let / else if let`, killing the `clippy::collapsible_else_if` from the other thread without an explicit lint suppression.

Force-send semantics on the Unavailable branch are preserved through *absence* of `continue` — when the rate-limit window has elapsed (or no record exists), we fall through to the single send path below.

Proposed shape:

```rust
async fn ensure_pool_validator_sets_updated(&mut self) -> anyhow::Result<bool> {
    let provider = contract_provider!(self.rpc_client.clone());
    let mut all_updated = true;
    let now = time_format::now();

    // Fetch once per tick. None → fall back to per-pool rate-limited send.
    let current_vset_hash: Option<[u8; 32]> =
        match self.rpc_client.get_config_param_cell(34).await {
            Ok(cell) => {
                let mut hash = [0u8; 32];
                hash.copy_from_slice(cell.repr_hash().as_slice());
                Some(hash)
            }
            Err(e) => {
                tracing::warn!(
                    target: "contracts",
                    "update_validator_set: config param 34 unavailable, proceeding with rate-limited fallback: {:#}",
                    e
                );
                None
            }
        };

    for (node_id, pool_binding) in
        self.pools.iter().filter(|(_, b)| b.pool_kind() == PoolKind::TONCore)
    {
        let validator_wallet = match self.wallets.get(node_id.as_str()) {
            Some(w) => (*w).clone(),
            None => {
                tracing::warn!(target: "contracts",
                    "[{}] ensure_pool_validator_sets_updated: no validator wallet in config (skip TonCore pools)",
                    node_id);
                all_updated = false;
                continue;
            }
        };
        let wallet_addr = validator_wallet.address().await?;
        let mut seqno: Option<i64> = None;

        for pool in pool_binding.inner_pools() {
            let pool_addr = pool.address().await?;
            let fallback_key = fallback_update_key(node_id, &pool_addr);
            let pool_data = match pool.get_pool_data().await {
                Ok(d) => d,
                Err(e) => {
                    tracing::warn!(target: "contracts",
                        "[{}] get_pool_data error (skipping update_validator_set): pool={} {:#}",
                        node_id, pool_addr, e);
                    continue;
                }
            };

            tracing::info!(target: "contracts",
                "[{}] pool={} state={} vsc_count={}",
                node_id, pool_addr, pool_data.state, pool_data.validator_set_changes_count);

            if pool_data.state != 2 || pool_data.validator_set_changes_count >= 2 {
                continue;
            }

            // Decide skip vs. send. Falling out of this block means "send".
            if let Some(current) = current_vset_hash {
                if current == pool_data.saved_validator_set_hash {
                    tracing::debug!(target: "contracts",
                        "[{}] skip update_validator_set: pool={}, vsc_count={}, validator set unchanged",
                        node_id, pool_addr, pool_data.validator_set_changes_count);
                    continue;
                }
                // hash differs → fall through
            } else if let Some(last_sent_at) =
                self.fallback_update_sent_at.get(&fallback_key).copied()
            {
                if now.saturating_sub(last_sent_at) < UPDATE_VALIDATOR_SET_FALLBACK_INTERVAL_SEC {
                    tracing::debug!(target: "contracts",
                        "[{}] skip fallback update_validator_set: pool={}, vsc_count={}, sent {}s ago",
                        node_id, pool_addr, pool_data.validator_set_changes_count,
                        now.saturating_sub(last_sent_at));
                    all_updated = false;
                    continue;
                }
                // window elapsed → fall through (force send)
            }
            // else: None and no record → fall through (first fallback send)

            let current_seqno = match seqno {
                Some(s) => s,
                None => provider
                    .get_method(wallet_addr.to_string(), "seqno", vec![])
                    .await?
                    .i64(0)?,
            };

            tracing::info!(target: "contracts",
                "[{}] update_validator_set: pool={}, state={}, vsc_count={}, from_wallet={}",
                node_id, pool_addr, pool_data.state,
                pool_data.validator_set_changes_count, wallet_addr);

            let body = tc_messages::update_validator_set(0)?;
            let msg = validator_wallet
                .build_message(pool_addr.clone(), POOL_OP_GAS, body, true,
                    Some(u32::try_from(current_seqno)?), None, None)
                .await?;
            self.broadcast(&msg).await?;
            // Record send time unconditionally — useful both for the fallback rate limit
            // and for de-duplicating Available-branch sends while the previous op6 is still in flight.
            self.fallback_update_sent_at.insert(fallback_key, time_format::now());
            seqno = Some(current_seqno + 1);
            all_updated = false;
        }
    }
    Ok(all_updated)
}
```

The `enum VsetHashState` (and its `Clone, Copy` derive) can be removed. Field name `fallback_update_sent_at` is now slightly off — it tracks every op6 send, not just fallback — consider renaming to `last_update_sent_at` and the helper to `update_send_key`. Optional, but it documents the unified semantics.

This also supersedes the `collapsible_else_if` comment above and partially the staleness comment (`now` for the rate-limit check still wants to be the per-call snapshot, but the insert uses a fresh `time_format::now()` right after broadcast).

mrnkslv: done

src/node-control/service/src/contracts/contracts_task.rs
mrnkslv: ok)

src/node-control/service/src/contracts/contracts_task.rs
mrnkslv: Acknowledged — keeping hardcoded for now; will add to ContractsAutomationConfig when a concrete tuning need arises.

src/node-control/service/src/contracts/contracts_task.rs
mrnkslv: ok

src/node-control/service/src/contracts/contracts_task.rs
mrnkslv: noted for future

src/node-control/service/src/contracts/contracts_task.rs
mrnkslv: done

src/node-control/service/src/contracts/contracts_task.rs
mrnkslv: done

src/node-control/service/src/contracts/contracts_task.rs
mrnkslv: done

Keshoid (approved):

PR #194, opened 2026-05-28
Feature/sma 101 audit log trait noop factory composition root wiring

mrnkslv:
## Summary

Define AuditLog async trait, NoopAuditLog, and AuditLogFactory as the foundation for the audit log subsystem. Wire Arc<dyn AuditLog> through the composition root so all future producers can depend on the trait without further structural changes.

## Changes

- Add AuditLog async trait (record(&self, event: AuditEvent)) and NoopAuditLog implementation in audit/log.rs
- Add AuditLogFactory::from_config returning Arc<dyn AuditLog>; placeholder returns NoopAuditLog until JsonlAuditLog is implemented
- Add AuditInitError using thiserror with InvalidPath, DirCreate, FileOpen variants
- Add AuditLogConfig to AppConfig with #[serde(default)] — existing configs without the field load unchanged
Wire Arc<dyn AuditLog> in service_main_task.rs: initialized after setup_log, passed to ElectionsTask and http_server_task::run
- Add audit: Arc<dyn AuditLog> field to AppState and ElectionsTask
- Remove standalone audit/config.rs — AuditLogConfig lives in common::app_config alongside other config types
- Tests: noop_audit_log_record_completes verifies the NoopAuditLog contract;
- factory_returns_audit_log_that_accepts_events verifies factory behavior without downcasting

Closes SMA-101


src/node-control/service/src/audit/mod.rs
```diff
@@ -6,15 +6,18 @@
...
  */
-pub mod config;
 pub mod enums;
 pub mod event;
+pub mod factory;
+pub mod log;
 pub mod participant;
 
-pub use config::AuditLogConfig;
+pub use common::app_config::AuditLogConfig;
```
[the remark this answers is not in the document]
mrnkslv: done

src/node-control/service/src/service_main_task.rs
```diff
@@ -52,6 +53,11 @@ pub async fn run_with_config(
         .await
         .context("initialize runtime config store")?;
     let runtime_cfg = Arc::new(runtime_cfg);
+
+    let audit = AuditLogFactory::from_config(&app_cfg.audit_log)
+        .await
+        .map_err(|e| anyhow::anyhow!("audit log init failed: {e}"))?;
```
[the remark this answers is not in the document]
mrnkslv: done

src/node-control/service/src/audit/factory.rs
```diff
@@ -0,0 +1,83 @@
...
+    AuditLogConfig,
+    log::{AuditLog, NoopAuditLog},
+};
+use std::sync::Arc;
+use thiserror::Error;
+
+pub struct AuditLogFactory;
+
+impl AuditLogFactory {
+    pub async fn from_config(config: &AuditLogConfig) -> Result<Arc<dyn AuditLog>, AuditInitError> {
```
Keshoid: **Missing `enabled` flag deviates from the SMA-101 spec.**

The Linear ticket explicitly defines:

```
if !config.enabled {
    return Ok(Arc::new(NoopAuditLog));
}
unimplemented!("JsonlAuditLog wiring is implemented in SMA-99.3")
```

The current implementation drops `enabled` entirely and unconditionally returns `NoopAuditLog`, which has two consequences:

1. The `factory_panics_when_enabled_before_99_3` safety-net test (also in the spec) silently disappears. That test is what guarantees SMA-99.3 cannot be forgotten — once `JsonlAuditLog` lands, nothing forces it to actually be wired.
2. Operators lose a kill-switch for audit logging once the real writer is in place (e.g. during disk-pressure incidents or triage).

Suggest adding `pub enabled: bool` (default `true`) to `AuditLogConfig` and restoring the spec's branching here, including the `unimplemented!()` arm and the corresponding test.

src/node-control/service/src/elections/election_task.rs
```diff
@@ -29,6 +29,7 @@ pub async fn run(
     runtime_cfg: Arc<dyn RuntimeConfig>,
     store: Arc<SnapshotStore>,
     on_status_change: Option<BindingStatusCallback>,
+    _audit: Arc<dyn AuditLog>,
```
Keshoid: **`audit` is accepted but never threaded into `ElectionRunner`.**

`service_main_task.rs` passes `audit.clone()` to `ElectionsTask::new`, but here it's named `_audit` and dropped on the floor. The spec specifically calls for `ElectionRunner::new(/* ..., */ audit.clone())` — the whole point of this PR is "future producers can depend on the trait without further structural changes," and skipping the runner means SMA-99.4 will re-do this plumbing.

Even if no producer call sites exist yet, please thread it down to `ElectionRunner` so the field is in place. Otherwise this is half a wire-up.

src/node-control/service/src/audit/factory.rs
```diff
@@ -0,0 +1,83 @@
...
+        // SMA-99.3: spawn JsonlAuditLog writer from `config`.
```
Keshoid: Nit: prefer `_config: &AuditLogConfig` in the signature over `let _ = config;` inside the body. Reads more clearly as "intentionally unused for now."

Keshoid (commented):
Spec compliance review for SMA-101. Three issues worth addressing before merge — two medium (deviation from acceptance criteria), one nit.

## Acceptance criteria check

| Criterion | Status |
|---|---|
| `Arc<dyn AuditLog>` built in composition root and passed to elections/http | Partial — passed to `election_task::run` but not into `ElectionRunner::new` |
| With `enabled = false`, service uses `NoopAuditLog` | `enabled` field omitted from `AuditLogConfig` |
| With `enabled = true`, factory returns `unimplemented!` | No longer applicable — factory always returns Noop |

## What's good

- `#[async_trait]` shape and `record(&self, AuditEvent)` match the spec.
- `#[serde(default)]` on `audit_log` + the `app_config_deserializes_without_audit_log_field` test cover existing-config migration cleanly.
- Move of `AuditLogConfig` into `common::app_config` consolidates config types and deletes the duplicate (per earlier Copilot feedback).
- `service_main_task.rs` switched to `.context(...)` so the `AuditInitError` source chain survives.
- CI workflow change carries a self-documenting `delete after merge into release branch` comment.

See inline comments for the two spec deviations and one nit.

src/node-control/service/src/audit/factory.rs
mrnkslv: we discussed and decided, that we don't need "enable" param, because it's always will be in true state)

src/node-control/service/src/audit/factory.rs
mrnkslv: done

src/node-control/service/src/elections/election_task.rs
mrnkslv: done

Keshoid (approved):

PR #198, opened 2026-06-01
Feature/sma 102 audit log jsonlauditlog writer task batching rotation fsync

mrnkslv:
## Summary

Replace the NoopAuditLog placeholder in AuditLogFactory with a production
JsonlAuditLog implementation. Audit events are serialized to JSON Lines,
written asynchronously by a background task with batching and file rotation,
and flushed deterministically during service shutdown.

## Changes

- Add `JsonlAuditLog` as the producer-facing `AuditLog` implementation:
  non-blocking `record()` via MPSC queue, dropped-event counter, and
  configurable backpressure timeout.
- Add `AuditWriter` background task:
  - time- and size-based batching;
  - size-based rotation with `max_files` retention;
  - synthetic `system.audit_events_dropped` recovery events;
  - Unix file mode 0o600.
- Extend the `AuditLog` trait with a default `shutdown()` hook so the
  composition root can drain pending events through `Arc<dyn AuditLog>`.
- Wire `JsonlAuditLog::start` into `AuditLogFactory::from_config` and call
  `audit.shutdown().await` from `service_main_task` after all producers stop.
- Harden rotation and shutdown behavior:
  - guard `max_files = 0` (no underflow);
  - treat `max_files = 1` as “live file only” (truncate in place, no history);
  - close the file handle before rename (portable across platforms, no `/dev/null` trick);
  - use async `try_exists` instead of blocking `Path::exists`.
- Improve operability: include `event_id`, `source`, and `subject` in the
  tracing warning when an event is dropped due to a full queue.
- Add unit tests for batching, rotation, retention limits, queue-full drops,
  dropped-event recovery, concurrent writes, shutdown flush, crash-window
  recovery on restart, and Unix file permissions.
  
  closes SMA-102


.github/workflows/ci.yml
```diff
@@ -2,7 +2,10 @@ name: CI
 
 on:
   pull_request:
-    branches: ["master", "release/**"]
+    branches:
+      - master
+      - release/**
+      - feature/sma-99-audit-log-architecture  # delete after merge into release branch
```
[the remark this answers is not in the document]
mrnkslv: we'll delete that before merge into release branch

src/node-control/service/src/audit/jsonl_log.rs
```diff
@@ -0,0 +1,132 @@
...
+
+    async fn start_inner(
+        config: AuditLogConfig,
+        write_delay: Duration,
+    ) -> Result<Arc<Self>, AuditInitError> {
+        let config = Arc::new(config);
+        std::fs::create_dir_all(config.path.parent().ok_or_else(|| {
+            AuditInitError::InvalidPath(config.path.to_string_lossy().to_string())
+        })?)
+        .map_err(AuditInitError::DirCreate)?;
```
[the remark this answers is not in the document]
mrnkslv: accepted

src/node-control/service/src/audit/jsonl_log.rs
```diff
@@ -0,0 +1,132 @@
...
+
+        let (tx, rx) = mpsc::channel(config.queue_capacity);
```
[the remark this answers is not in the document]
mrnkslv: done

src/node-control/service/src/audit/jsonl_writer.rs
```diff
@@ -0,0 +1,680 @@
...
+            dropped_events: dropped,
+            last_dropped_seen: 0,
+            write_delay,
+        })
+    }
+
+    pub(crate) async fn run(mut self, mut rx: mpsc::Receiver<AuditCommand>) {
+        let mut interval =
+            tokio::time::interval(Duration::from_millis(self.config.batch_interval_ms));
+        interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Delay);
```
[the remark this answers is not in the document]
mrnkslv: +

src/node-control/service/src/audit/jsonl_writer.rs
```diff
@@ -0,0 +1,680 @@
...
+            tokio::fs::rename(path, &rotated).await?;
+            opts.append(true).create(true);
+        } else {
+            // No history retained: truncate the live file in place.
+            opts.write(true).create(true).truncate(true);
+        }
+        #[cfg(unix)]
+        opts.mode(0o600);
+        self.file = Some(opts.open(path).await?);
+        self.current_size = 0;
```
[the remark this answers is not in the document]
mrnkslv: +

src/node-control/service/src/audit/jsonl_log.rs
```diff
@@ -0,0 +1,135 @@
...
+    }
+}
+
+#[async_trait]
+impl AuditLog for JsonlAuditLog {
+    async fn shutdown(&self) {
+        // Signal the writer to drain and flush the final batch. The writer also
+        // stops on channel close, but sending an explicit command lets us await
+        // the flush deterministically.
+        let _ = self.sender.send(AuditCommand::Shutdown).await;
```
Keshoid: **[HIGH] `shutdown()` can block indefinitely on a full queue or hung writer.**

`send(AuditCommand::Shutdown).await` has no timeout — if the writer is stuck on slow IO and the channel is at capacity, this send never returns, and the entire `service_main_task::run_with_config` shutdown hangs (there's no timeout around `audit.shutdown().await` either — see `service_main_task.rs:174`).

Suggest wrapping with `tokio::time::timeout(...)` and falling back to `handle.abort()` on timeout, or routing `Shutdown` through a separate `oneshot` so it cannot be blocked by a full event channel.

src/node-control/service/src/audit/jsonl_writer.rs
```diff
@@ -0,0 +1,755 @@
...
+                    Self::io_failed("audit write before rotation failed", e);
+                }
+                if let Err(e) = self.rotate().await {
+                    Self::io_failed("audit rotation failed", e);
+                    continue;
+                }
+            }
+            self.batch.extend_from_slice(&line);
+            self.batch.push(b'\n');
+            self.current_size += needed as u64;
```
Keshoid: **[HIGH] `current_size` accounting drifts on write failure.**

The counter is bumped after appending each event to `self.batch` but *before* `write_batch_and_clear()` actually persists the bytes. If the final write at line 153 fails, `io_failed` only logs and returns — `self.batch` is not cleared (it's cleared inside `write_batch_and_clear` only on success), and the next `flush()` clears it unconditionally at line 129.

Net effect: bytes are lost on disk, but `current_size` has already counted them. Subsequent rotation checks fire prematurely.

Fix: either clear `self.batch` in the failure branch of `write_batch_and_clear`, or only bump `current_size` after the successful write.

src/node-control/service/src/audit/jsonl_writer.rs
```diff
@@ -0,0 +1,755 @@
...
+    async fn reopen_live_append(&mut self, path: &std::path::Path) -> std::io::Result<()> {
+        let mut opts = tokio::fs::OpenOptions::new();
+        opts.append(true).create(true);
+        #[cfg(unix)]
+        opts.mode(0o600);
+        let file = opts.open(path).await?;
+        self.current_size = file.metadata().await?.len();
+        self.file = Some(file);
+        Ok(())
+    }
```
Keshoid, line 332: **[HIGH] `reopen_live_append` skips the `set_permissions(0o600)` belt-and-suspenders.**

The initial `open()` (lines 62-74) does mode-on-create *plus* a follow-up `tokio::fs::set_permissions(path, 0o600)`. This recovery path only sets `opts.mode(0o600)`, which `OpenOptions` silently ignores for an already-existing file. If the live file is recreated after a failed rotation and happens to already exist with wider permissions, those permissions stay.

Mirror the `open()` pair here.

src/node-control/service/src/audit/jsonl_log.rs
```diff
@@ -0,0 +1,135 @@
...
+                            ?source,
+                            ?subject,
+                            "audit event dropped: queue full after timeout"
+                        );
+                    }
+                }
+            }
+            Err(mpsc::error::TrySendError::Closed(_)) => {
+                tracing::error!("audit log channel closed; service likely shutting down");
+            }
```
Keshoid, line 181: **[HIGH] Silent drop on `TrySendError::Closed` — `dropped_events` not incremented.**

The event is dropped and the error is logged, but the counter doesn't move — so no synthetic `system.audit_events_dropped` event will ever be emitted for these losses. If the channel-closed window is non-trivial (race with shutdown, writer crashed and rx dropped, …) we get a gap in the audit trail with no upcoming-marker.

Suggest `self.dropped_events.fetch_add(1, Ordering::Relaxed)` here as well, so the recovery event covers both timeout-drops and closed-drops.

Keshoid (commented):
Second-pass review. Overall the writer is solid — bounded backpressure, typed payloads, crash-window test, portable rotation. Below are correctness, reliability, and style findings inline. Severity tags: **[HIGH]** needs fixing before merge; **[MED]** reliability or testability concern; **[LOW]** nit / micro-opt.

src/node-control/service/src/audit/jsonl_log.rs
```diff
@@ -0,0 +1,135 @@
...
+        // stops on channel close, but sending an explicit command lets us await
+        // the flush deterministically.
+        let _ = self.sender.send(AuditCommand::Shutdown).await;
+
+        // Take the handle without holding the lock across the await.
+        let handle = self.writer.lock().expect("audit writer lock poisoned").take();
+        if let Some(handle) = handle {
+            let _ = handle.await;
+        }
+    }
```
Keshoid, line 153: **[MED] Race between concurrent `shutdown()` calls.**

If two tasks call `shutdown()` concurrently: one takes the handle and awaits it; the other gets `None` from `.take()` and returns immediately — *before* the writer has finished draining. The contract "after `shutdown()` returns, the writer is done" is broken for the second caller.

Two simple fixes:
- a `tokio::sync::OnceCell<()>` set by the first completer that all callers `.get_or_init`/`.await` on;
- a `tokio::sync::Notify` that all callers `.notified()` on, and the first completer calls `.notify_waiters()` after the join.

src/node-control/service/src/audit/jsonl_writer.rs
```diff
@@ -0,0 +1,755 @@
...
+        let file = self
+            .file
+            .as_mut()
+            .ok_or_else(|| std::io::Error::other("audit file handle not open"))?;
+        file.write_all(&self.batch).await?;
+        if self.config.fsync_on_batch {
+            file.sync_data().await?;
+        }
+        #[cfg(test)]
+        file.sync_all().await?;
```
Keshoid: **[MED] `#[cfg(test)] file.sync_all().await?` breaks test/prod isolation.**

Tests always fsync regardless of `fsync_on_batch`, which (a) means test behavior differs from production and (b) hides bugs where production code forgets fsync when `fsync_on_batch = true`.

`run_writer_session` already serializes producer and writer through `tokio::join!` — an fsync here shouldn't be necessary for determinism. If there's residual flakiness, the root cause likely isn't sync (see the comment about the global `WRITER_TEST_LOCK`).

src/node-control/service/src/audit/jsonl_writer.rs
```diff
@@ -0,0 +1,755 @@
...
+    static WRITER_TEST_LOCK: std::sync::Mutex<()> = std::sync::Mutex::new(());
+
+    async fn run_writer_test<F, Fut>(f: F)
+    where
+        F: FnOnce() -> Fut,
+        Fut: std::future::Future<Output = ()>,
+    {
+        let _guard = WRITER_TEST_LOCK.lock().unwrap_or_else(std::sync::PoisonError::into_inner);
+        f().await;
+    }
```
Keshoid, line 446: **[MED] `WRITER_TEST_LOCK` global mutex hides an unidentified race.**

Tests use `flavor = "current_thread"` and isolated `tempdir()`s — they should be naturally independent. The comment cites "Tokio blocking pool contention" as a hypothesis, not a root cause. If each test on its own runtime with its own tempdir still races, the bug is in the writer itself or in `run_writer_session`.

Worth one more debug pass — a global lock kills test parallelism and masks the real problem (and compounds the `sync_all` smell above).

src/node-control/service/src/audit/jsonl_writer.rs
```diff
@@ -0,0 +1,755 @@
...
+
+    #[cfg(test)]
+    fn io_failed(context: &str, err: std::io::Error) {
+        panic!("{context}: {err}");
+    }
+
+    #[cfg(not(test))]
+    fn io_failed(context: &str, err: std::io::Error) {
+        tracing::error!(error = %err, "{context}");
+    }
```
Keshoid: **[MED] `Self::io_failed` panics under `#[cfg(test)]`, leaving the rotation-recovery path untestable.**

The main new feature in this PR is `reopen_live_append` as a fallback when rotation fails. To unit-test it we need to inject an IO failure inside `rotate_inner` and observe that the writer recovered — but `io_failed` then panics the test before any assertion can run.

Suggest making the test-mode behavior match prod (`tracing::error!`) and using an external counter / `Notify` / channel for assertions in tests that exercise the recovery path.

src/node-control/service/src/audit/jsonl_writer.rs
```diff
@@ -0,0 +1,755 @@
...
+    async fn rotate(&mut self) -> std::io::Result<()> {
+        let path = self.config.path.clone();
+        // Total retained files (including the live one) is at least 1; the
+        // number of rotated history segments is `max - 1`. Guarding against 0
+        // avoids an arithmetic underflow on `max - 1`.
+        let max = self.config.max_files.max(1);
+
+        if max > 1 {
+            // Remove the oldest segment (.{max-1}) if present.
+            let oldest = path.with_extension(format!("jsonl.{}", max - 1));
```
Keshoid: **[MED] Rotation paths assume the live file ends in `.jsonl`.**

`path.with_extension("jsonl.N")` *replaces* the existing extension. If an operator configures `path = "/var/log/audit.log"`, the live file is `audit.log` but rotated files become `audit.jsonl.1`, `audit.jsonl.2`, … — silent drift, and `count_rotated_files` (in tests) wouldn't even find them.

Either validate the config at start (require `.jsonl` suffix) or build rotated names via `path.with_file_name(format!("{}.{n}", path.file_name().unwrap().to_string_lossy()))`. Same concern at line 246.

src/node-control/service/src/audit/jsonl_writer.rs
```diff
@@ -0,0 +1,755 @@
...
+        #[cfg(unix)]
+        opts.mode(0o600);
+        let file = opts.open(path).await.map_err(AuditInitError::FileOpen)?;
+
+        #[cfg(unix)]
+        {
+            use std::os::unix::fs::PermissionsExt;
+            tokio::fs::set_permissions(path, std::fs::Permissions::from_mode(0o600))
+                .await
+                .map_err(AuditInitError::FileOpen)?;
```
Keshoid: **[MED] `set_permissions` failure maps to `AuditInitError::FileOpen` — semantic mismatch.**

An operator reading "failed to open audit file" when the file actually opened fine but `chmod` failed will be misled. Same for the `metadata().len()` failure on line 76 (used to initialize `current_size`, also mapped to `FileOpen`).

Suggest distinct variants like `AuditInitError::SetPermissions(#[source] io::Error)` and `AuditInitError::Metadata(#[source] io::Error)` so logs point at the real failure.

src/node-control/service/src/audit/jsonl_writer.rs
```diff
@@ -0,0 +1,755 @@
...
+        let mut opts = tokio::fs::OpenOptions::new();
+        if max > 1 {
+            // Preserve history: rename current -> .1, then open a fresh live file.
+            let rotated = path.with_extension("jsonl.1");
+            tokio::fs::rename(path, &rotated).await?;
+            opts.append(true).create(true);
+        } else {
+            // No history retained: truncate the live file in place.
+            opts.write(true).create(true).truncate(true);
+        }
```
Keshoid, line 385: **[LOW] `max_files = 1` truncates the live file in place — breaks live-tail consumers.**

`opts.write(true).create(true).truncate(true)` on the active file means any external consumer tailing it (`tail -f`, Vector, journalbeat, Loki promtail) sees "file truncated" mid-stream and loses position.

Worth documenting for operators in the README, or rejecting `max_files = 1` outright in prod configs. The `max_files > 1` append+rename path preserves tail-compatibility because the inode the consumer has open continues to receive writes until the rename.

src/node-control/service/src/audit/jsonl_writer.rs
```diff
@@ -0,0 +1,755 @@
...
+        interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Delay);
+        let mut buffered: Vec<AuditEvent> = Vec::with_capacity(self.config.batch_max_events);
+
+        loop {
+            tokio::select! {
+                cmd = rx.recv() => {
+                    match cmd {
+                        Some(AuditCommand::Event(ev)) => {
+                            buffered.push(*ev);
+                            if buffered.len() >= self.config.batch_max_events {
```
Keshoid: **[LOW] `batch_max_events = 0` ⇒ flush on every event.**

Not a crash — `if buffered.len() >= 0` is always true, so each individual event triggers a flush plus a write syscall. Worth clamping `batch_max_events` to `>= 1` in `start_inner` (or rejecting `0` in config validation).

src/node-control/service/src/audit/jsonl_writer.rs
```diff
@@ -0,0 +1,755 @@
...
+        for ev in buffered.drain(..) {
+            let line = match serde_json::to_vec(&ev) {
+                Ok(v) => v,
+                Err(e) => {
+                    tracing::error!(error = %e, "failed to serialize audit event");
+                    continue;
+                }
+            };
+            let needed = line.len() + 1;
+            if self.current_size + needed as u64 > self.config.max_size_bytes {
```
Keshoid: **[LOW] `max_size_bytes = 0` ⇒ rotation on every event.**

With `max_size_bytes = 0`, the predicate `current_size + needed > 0` is always true on the first event. After rotation `current_size = 0`, but the very next event triggers another rotation. Effectively: one rotation per event indefinitely. Worth clamping in `start_inner`.

src/node-control/service/src/audit/jsonl_writer.rs
```diff
@@ -0,0 +1,755 @@
...
+                if let Err(e) = self.write_batch_and_clear().await {
+                    Self::io_failed("audit write before rotation failed", e);
+                }
+                if let Err(e) = self.rotate().await {
+                    Self::io_failed("audit rotation failed", e);
+                    continue;
+                }
+            }
+            self.batch.extend_from_slice(&line);
+            self.batch.push(b'\n');
```
Keshoid: **[LOW] `serde_json::to_vec(&ev)` then `extend_from_slice(&line)` allocates a fresh `Vec<u8>` per event only to copy it into `self.batch`.**

`serde_json::to_writer(&mut self.batch, &ev)?` writes the event directly into the reused batch buffer (and you still get the same `serde_json::Error` to log). Then a single `self.batch.push(b'\n')`. Micro-perf but free.

src/node-control/service/src/audit/jsonl_log.rs
```diff
@@ -0,0 +1,135 @@
...
+    async fn shutdown(&self) {
+        // Signal the writer to drain and flush the final batch. The writer also
+        // stops on channel close, but sending an explicit command lets us await
+        // the flush deterministically.
+        let _ = self.sender.send(AuditCommand::Shutdown).await;
+
+        // Take the handle without holding the lock across the await.
+        let handle = self.writer.lock().expect("audit writer lock poisoned").take();
+        if let Some(handle) = handle {
+            let _ = handle.await;
```
Keshoid: **[LOW] `let _ = handle.await` swallows panic info from the writer task.**

If the writer panicked (e.g. inside `flush` from a serializer or filesystem invariant), we won't know.

```rust
if let Err(e) = handle.await {
    tracing::error!(error = %e, "audit writer task panicked");
}
```

src/node-control/service/src/service_main_task.rs
```diff
@@ -168,5 +168,9 @@ pub async fn run_with_config(
         let _ = task.disable().await;
     }
     let _ = http_task_handle.await;
```
Keshoid: **[LOW] Same pattern as the audit-writer await: `let _ = http_task_handle.await` swallows panic info from the HTTP server task.**

Log the join error so HTTP-task panics aren't silent during shutdown.

src/node-control/service/src/audit/jsonl_log.rs
```diff
@@ -0,0 +1,135 @@
...
+
+        match self.sender.try_send(cmd) {
+            Ok(()) => return,
+            Err(mpsc::error::TrySendError::Full(cmd)) => {
+                // Capture diagnostics before `cmd` is moved into the send future,
+                // so a dropped event can be attributed to its source/subject.
+                let diag = match &cmd {
+                    AuditCommand::Event(ev) => Some((ev.id, ev.source, ev.subject.kind)),
+                    _ => None,
+                };
```
Keshoid: **[LOW] The `_ => None` arm is unreachable.**

`cmd` was just constructed as `AuditCommand::Event(Box::new(event))` on line 101 and hasn't been moved since. An `if let AuditCommand::Event(ev) = &cmd { ... }` (or a direct field read before constructing `cmd`) is clearer and removes the dead arm.

src/node-control/service/src/audit/jsonl_writer.rs
```diff
@@ -0,0 +1,755 @@
...
+        let event = AuditEvent {
+            schema_version: 1,
+            id: Uuid::new_v4(),
+            ts: Utc::now(),
+            source: AuditSource::System,
+            severity: AuditSeverity::Warn,
+            outcome: AuditOutcome::Failure,
+            actor: AuditActor { kind: AuditActorKind::System, id: None, role: None, ip: None },
+            subject: AuditSubject {
+                kind: AuditSubjectKind::Config,
```
Keshoid: **[LOW] `AuditSubjectKind::Config` for `system.audit_events_dropped` is semantically wrong.**

The synthetic recovery event isn't about a configuration object. The inline `// closest match; or add AuditSubjectKind::System` comment already flags this.

Tied to the broader event-format redesign discussion — fix together with adding `AuditSubjectKind::System` (or the enum-based `AuditTarget` proposal from the format thread).

src/node-control/service/src/audit/jsonl_log.rs
```diff
@@ -0,0 +1,135 @@
...
+        }
+
+        let (tx, rx) = mpsc::channel(config.queue_capacity.max(1));
+        let dropped_events = Arc::new(AtomicU64::new(0));
+
+        let writer = AuditWriter::open(config.clone(), dropped_events.clone(), write_delay).await?;
+        let handle = tokio::spawn(writer.run(rx));
+
+        Ok(Arc::new(Self { sender: tx, dropped_events, config, writer: Mutex::new(Some(handle)) }))
+    }
```
Keshoid, line 112: **[LOW] `start_with_write_delay` + `write_delay` parameter pollutes the production constructor signature with a test-only hook.**

With the public `start_inner(config, write_delay)`, every reader of `JsonlAuditLog` has to wonder what `write_delay` is for. Some isolation options:

- a `WriterDeps` trait that production and tests implement (test impl sleeps);
- a test-only `AuditWriter::with_delay` factory behind `#[cfg(test)]` that bypasses `start`;
- if the only goal is to slow IO down deterministically, a `tokio::time::pause()` + manual `advance` in the test removes the need for any artificial delay altogether.

Not blocking, but it's a visible test smell on the public producer surface.

src/node-control/service/src/audit/jsonl_writer.rs
```diff
@@ -0,0 +1,755 @@
...
+
+pub(crate) enum AuditCommand {
+    Event(Box<AuditEvent>),
+    Shutdown,
+
+    /// Forces an immediate flush of buffered events. Used only by tests to make
+    /// assertions deterministic without waiting for the batch interval.
+    #[cfg(test)]
+    Flush,
+}
```
Keshoid, line 45: **[LOW] `AuditCommand` has no `Debug` derive.**

Diagnostic logging (`tracing::warn!(?cmd, ...)`) isn't possible without it. `AuditEvent`, `AuditCommand`, and the inner types are all Debug-able, so adding `#[derive(Debug)]` here is free.

src/node-control/service/src/audit/jsonl_writer.rs
```diff
@@ -0,0 +1,755 @@
...
+            last_dropped_seen: 0,
+            write_delay,
+        })
+    }
+
+    pub(crate) async fn run(mut self, mut rx: mpsc::Receiver<AuditCommand>) {
+        let interval_ms = self.config.batch_interval_ms.max(1);
+        let mut interval = tokio::time::interval(Duration::from_millis(interval_ms));
+        interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Delay);
+        let mut buffered: Vec<AuditEvent> = Vec::with_capacity(self.config.batch_max_events);
```
Keshoid: **[LOW] First `interval.tick().await` fires immediately at t=0.**

That's the default `tokio::time::interval` behavior — `MissedTickBehavior::Delay` controls catch-up after misses, not the initial tick. On startup we therefore run an empty `flush` + empty `maybe_emit_dropped_recovery` immediately. Harmless but a wasted wakeup. Either `interval_at(Instant::now() + dur, dur)` or consume the first tick before entering the loop.

src/node-control/service/src/audit/jsonl_log.rs
```diff
@@ -0,0 +1,135 @@
...
+        config: AuditLogConfig,
+        write_delay: Duration,
+    ) -> Result<Arc<Self>, AuditInitError> {
+        let config = Arc::new(config);
+        if let Some(parent) = config.path.parent() {
+            if !parent.as_os_str().is_empty() {
+                tokio::fs::create_dir_all(parent).await.map_err(AuditInitError::DirCreate)?;
+            }
+        } else {
+            return Err(AuditInitError::InvalidPath(config.path.to_string_lossy().to_string()));
```
Keshoid: **[LOW] `AuditInitError::InvalidPath` is practically unreachable.**

`Path::parent()` returns `None` only for paths that terminate in a root or prefix (e.g. `/`, `C:\`). For any normal config value the empty-string branch above handles it. Either remove the variant (and return `DirCreate` with a meaningful message) or extend it to actual path validation worth surfacing — e.g. require absolute paths in prod, require `.jsonl` suffix (see the rotation-suffix comment elsewhere).

src/node-control/service/src/audit/jsonl_log.rs
mrnkslv: accepted

src/node-control/service/src/audit/jsonl_writer.rs
mrnkslv: done

src/node-control/service/src/audit/jsonl_writer.rs
mrnkslv: done

src/node-control/service/src/audit/jsonl_log.rs
mrnkslv: fixed

src/node-control/service/src/audit/jsonl_log.rs
mrnkslv: done

src/node-control/service/src/audit/jsonl_writer.rs
mrnkslv: fixed

src/node-control/service/src/audit/jsonl_writer.rs
mrnkslv: done

src/node-control/service/src/audit/jsonl_writer.rs
mrnkslv: fixed

src/node-control/service/src/audit/jsonl_writer.rs
mrnkslv: ok

src/node-control/service/src/audit/jsonl_writer.rs
mrnkslv: ok

src/node-control/service/src/audit/jsonl_writer.rs
mrnkslv: ok

src/node-control/service/src/audit/jsonl_writer.rs
mrnkslv: done

src/node-control/service/src/audit/jsonl_writer.rs
mrnkslv: ok

src/node-control/service/src/audit/jsonl_writer.rs
mrnkslv: fixed

src/node-control/service/src/audit/jsonl_log.rs
mrnkslv: fixed

src/node-control/service/src/service_main_task.rs
mrnkslv: done

src/node-control/service/src/audit/jsonl_log.rs
mrnkslv: deleted

src/node-control/service/src/audit/jsonl_writer.rs
mrnkslv: done

src/node-control/service/src/audit/jsonl_log.rs
mrnkslv: fixed

src/node-control/service/src/audit/jsonl_writer.rs
mrnkslv: ✅

src/node-control/service/src/audit/jsonl_writer.rs
mrnkslv: ✅

src/node-control/service/src/audit/jsonl_log.rs
mrnkslv: done

Keshoid (approved):

PR #205, opened 2026-06-03
Feature/sma 103 audit log elections producers stake withdraw tick

mrnkslv:
Wire Arc<dyn AuditLog> into the elections runner and record structured
audit events at key lifecycle points. Events are written through the
existing JsonlAuditLog when audit_log.enabled is true (default in
generated config).
Elections producers:
- ElectionsKeyGenerated — after validator key is created
- ElectionsStakeSubmitted — after stake message is sent
- ElectionsStakeSkipped — on every skip path (excluded node, recover
  pending, pool not ready, withdraw queue, adaptive defer, insufficient
  funds, low balance on calc_stake/send_stake)
- ElectionsWithdrawProcessed / ElectionsWithdrawProcessFailed — TONCore
  process_withdraw_requests path
- ElectionsStakeRecovered — after recover stake BOC is sent
- ElectionsTickFailed — only when run() returns Err (no tick started/
  completed events; tracing covers successful ticks)
StakeSkipReason: replace Other with AdaptiveSleepPeriod,
AdaptiveWaitingPeriod, NodeExcluded, RecoverPending,
InsufficientStakeFunds. Add adaptive_split50_defer_reason() to classify
AdaptiveSplit50 defer without duplicating readiness checks in tests.
Implementation notes:
- Free functions elections_audit_emit / elections_audit_stake_* avoid
  borrow conflicts with &mut Node in participate()
- Withdraw/recover tx_hash uses message cell repr_hash, not full BOC hex
- calc_stake "not enough funds" error uses required= so parse_balance_error
  can populate required_nanotons in skip payload
Testing:
- InMemoryAuditLog test double (#[cfg(test)] only)
- TestHarness.audit + runner_tests for stake/withdraw/tick/subject
- round_trip_all_variants extended for new payload variants

Closes SMA-103


src/node-control/service/src/elections/runner.rs
```diff
@@ -360,6 +368,82 @@ impl ElectionRunner {
...
+                };
+            }
+            let min_stake = configs.elections_info.min_stake;
+            let fee = ELECTOR_STAKE_FEE + NPOOL_COMPUTE_FEE;
+            let pool_free = node.stake_balance(fee).await.unwrap_or(0);
+            return (
+                StakeSkipReason::InsufficientStakeFunds,
+                Some(min_stake),
+                Some(pool_free.saturating_add(elections_stake)),
+            );
```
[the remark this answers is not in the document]
mrnkslv: ✅

src/node-control/service/src/elections/runner.rs
```diff
@@ -874,12 +1016,37 @@ impl ElectionRunner {
...
+                    election_id,
+                    node_id,
+                    StakeSkipReason::LowWalletBalance,
+                    required,
+                    available,
+                )
+                .await;
+                return Err(e).context("stake calculation error");
+            }
+        };
```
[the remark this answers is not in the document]
mrnkslv: ✅

src/node-control/service/src/elections/runner.rs
```diff
@@ -937,7 +1105,33 @@ impl ElectionRunner {
...
+                        &audit,
+                        election_id,
+                        node_id,
+                        StakeSkipReason::LowWalletBalance,
+                        required,
+                        available,
+                    )
+                    .await;
+                    return Err(e);
+                }
```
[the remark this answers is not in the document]
mrnkslv: ✅

src/node-control/service/src/elections/runner.rs
```diff
@@ -963,16 +1157,42 @@ impl ElectionRunner {
...
+                                    &audit,
+                                    election_id,
+                                    node_id,
+                                    StakeSkipReason::LowWalletBalance,
+                                    required,
+                                    available,
+                                )
+                                .await;
+                                return Err(e);
+                            }
```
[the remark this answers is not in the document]
mrnkslv: fixed

src/node-control/service/src/elections/runner.rs
```diff
@@ -963,16 +1157,42 @@ impl ElectionRunner {
...
+                                election_id,
+                                node_id,
+                                StakeSkipReason::LowWalletBalance,
+                                required,
+                                available,
+                            )
+                            .await;
+                            return Err(e);
+                        }
+                        elections_audit_stake_submitted(&audit, node_id, node, stake).await;
```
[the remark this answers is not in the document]
mrnkslv: outdated

src/node-control/service/src/elections/runner.rs
```diff
@@ -1815,6 +2077,121 @@ impl ElectionRunner {
...
+    )
+    .await;
+}
+
+/// Parses `required=` / `available=` TON amounts from runner balance error strings.
+fn parse_balance_error(msg: &str) -> (Option<u64>, Option<u64>) {
+    fn tons_to_nanotons(fragment: &str) -> Option<u64> {
+        let tons: f64 = fragment.trim().parse().ok()?;
+        Some((tons * 1_000_000_000.0) as u64)
+    }
```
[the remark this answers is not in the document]
mrnkslv: done

src/node-control/service/src/audit/jsonl_log.rs
```diff
@@ -0,0 +1,135 @@
...
+                            None => (None, None, None),
+                        };
+                        tracing::warn!(
+                            ?event_id,
+                            ?source,
+                            ?subject,
+                            "audit event dropped: queue full after timeout"
+                        );
+                    }
+                }
```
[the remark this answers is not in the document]
mrnkslv: done

src/node-control/service/src/elections/runner.rs
```diff
@@ -1815,6 +2013,94 @@ impl ElectionRunner {
...
+            nanotons_to_dec_string(stake),
+            participant.max_factor,
+            node.stake_policy.to_string(),
+            submission_time,
+        ))
+        .await;
+}
+
+/// Parses `required=` / `available=` TON amounts from runner balance error strings.
+fn parse_balance_error(msg: &str) -> (Option<u64>, Option<u64>) {
```
Keshoid: **Major — audit signal coupled to free-form error strings.**

This function reverse-engineers `required=` / `available=` out of a `tracing`-style message. The PR even reworded `calc_stake`'s "not enough funds" error so this parser keeps working — that's backwards: the audit producer should hold the structured values, not recover them from text.

If anyone reformats a balance error (rename a variable, swap the order, lose the ` TON` suffix), `LowWalletBalance` events silently disappear. There's no test today that would catch that.

**Suggested fix:** make `calc_stake` / `send_stake` return a typed error, e.g.

```rust
enum CalcStakeError {
    InsufficientWalletBalance { required: u64, available: u64 },
    Other(anyhow::Error),
}
```

Then propagate `required` / `available` directly into the skip event and delete `parse_balance_error` + the error-wording constraint.

src/node-control/service/src/elections/adaptive_strategy.rs
```diff
@@ -67,6 +90,36 @@ pub(crate) fn is_adaptive_split50_ready(
     true
 }
 
+/// When [`is_adaptive_split50_ready`] would return `false`, reports which wait gate blocked.
+pub(crate) fn adaptive_split50_defer_reason(
```
Keshoid: **Major — defer-reason logic duplicated.**

`is_adaptive_split50_ready` (returns `bool`) and `adaptive_split50_defer_reason` (returns `Option<AdaptiveDeferReason>`) compute the same `sleep_deadline` / `wait_deadline` / `participants_count < min_validators` math. `calc_stake` calls the first, gets `false`, returns `(0, None)`; the runner then calls `classify_stake_zero` which calls the second. Two passes over identical logic, with obvious drift risk.

**Suggested fix:** collapse to one function:

```rust
pub(crate) fn adaptive_split50_status(...) -> Option<AdaptiveDeferReason>;
// `is_ready` becomes `.is_none()` at call sites.
```

Thread the reason through `AdaptiveStakeResult::Zero(AdaptiveStakeZero::Defer(reason))` so `classify_stake_zero` disappears entirely.

Keshoid (commented):
Reviewed against SMA-103 acceptance criteria. Producers are in the right places and the constructors / `InMemoryAuditLog` give a clean test surface. Four design-level concerns and six smaller items below — full write-up in chat / PR thread; this review captures the inline findings.

src/node-control/service/src/elections/runner.rs
```diff
@@ -1171,13 +1353,22 @@ impl ElectionRunner {
...
-                    .message(to_addr, RECOVER_FEE, Self::build_recover_stake_payload().await?)
-                    .await?,
-            )?;
+            let msg = node
+                .wallet
+                .message(to_addr, RECOVER_FEE, Self::build_recover_stake_payload().await?)
+                .await?;
+            let tx_hash = format!("{:x}", msg.repr_hash());
+            let msg_boc = write_boc(&msg)?;
             node.api.send_boc(&msg_boc).await?;
```
Keshoid: **Major — recover_stake send-failure is not audited.**

If `send_boc` fails here the error propagates to `run()` which records `last_error` and continues. No `Elections*` audit event is emitted, so a forensic reader sees successful recoveries but silent failures.

This is asymmetric with `process_pending_withdraw_requests`, which audits both `ElectionsWithdrawProcessed` and `ElectionsWithdrawProcessFailed`.

**Suggested fix:** add an `ElectionsStakeRecoverFailed { reason: String }` variant and emit it here on `Err`, mirroring the withdraw pair.

src/node-control/service/src/elections/runner.rs
```diff
@@ -1815,6 +2013,94 @@ impl ElectionRunner {
...
+            reason,
+            required_nanotons.map(nanotons_to_dec_string),
+            available_nanotons.map(nanotons_to_dec_string),
+        ))
+        .await;
+}
+
+/// Emits `LowWalletBalance` only when `err` matches the `required=` / `available=` format
+/// (calc_stake or send_stake balance guards). Other failures are not audited as skip.
+async fn elections_audit_low_balance_if_parsed(
```
Keshoid: **Major — `participate` failures are audited inconsistently.**

Every `send_stake` / `calc_stake` failure routes through this helper. If the error string matches `required=…/available=…` it becomes a `LowWalletBalance` skip event; otherwise nothing is emitted and the error only lands in logs / `last_error`.

The current design is "audit successes and skips; failures live in logs" — defensible, but then string-parsed balance errors should not be the one exception that smuggles a failure into the audit stream as a skip. Pick one:

- **Either** audit all `participate` failures as a new `ElectionsStakeFailed { reason }` (preferred for forensics), **or**
- drop the low-balance string parsing and accept that mid-flight failures stay in logs only.

The current half-and-half is the worst of both worlds and ties the audit surface to the error-message format.

src/node-control/service/src/elections/runner.rs
```diff
@@ -1135,18 +1293,42 @@ impl ElectionRunner {
             .await
             .context("build process_withdraw_requests message")?;
         let msg_boc = write_boc(&msg).context("encode process_withdraw_requests boc")?;
-        node.api.send_boc(&msg_boc).await.context("send process_withdraw_requests boc")?;
+        let tx_hash = format!("{:x}", msg.repr_hash());
```
Keshoid: **Minor — `tx_hash` is misleading.** This is the outbound message cell's `repr_hash`, not the on-chain transaction hash. Consumers will try `tonscan.org/tx/<hash>` and silently get nothing. The PR description acknowledges this.

Schema is still pre-rev-1, so rename now is cheap: `msg_hash` (or `outbound_msg_hash`) on both the payload field and the constructor parameter. Same applies at the recover_stake site below.

src/node-control/service/src/elections/runner.rs
```diff
@@ -360,6 +358,66 @@ impl ElectionRunner {
...
+            reason,
+            required_nanotons,
+            available_nanotons,
+        )
+        .await;
+    }
+
+    /// Classifies stake==0 when [`calc_stake`] did not already supply an adaptive reason
+    /// (defer windows only; adaptive zero cases come from [`calc_adaptive_stake`]).
+    async fn classify_stake_zero(
```
Keshoid: **Minor — `async` without an `await`.** This function never awaits — drop `async` (and the `.await` at the call site) to remove a needless future. (Becomes moot if Major-2 is taken; this function disappears.)

src/node-control/service/src/elections/runner.rs
```diff
@@ -360,6 +358,66 @@ impl ElectionRunner {
         }
     }
 
+    async fn audit_stake_skipped(
```
Keshoid: **Minor — dead duplication.** `audit_stake_skipped` (method on `&self`) is a thin wrapper over the free function `elections_audit_stake_skipped` introduced for the `&mut Node` borrow conflict in `participate`. Every call site already has `self.audit` reachable, so the method buys nothing — pick one and delete the other. Recommend keeping the free function (since it's the form required for the borrow workaround) and deleting this method.

src/node-control/service/src/audit/in_memory.rs
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * This software is provided "AS IS", WITHOUT WARRANTY OF ANY KIND.
+ */
+use crate::audit::{AuditEvent, log::AuditLog};
+use async_trait::async_trait;
+use std::sync::Mutex;
+
+/// Captures audit events in memory for unit tests.
+pub struct InMemoryAuditLog {
+    pub events: Mutex<Vec<AuditEvent>>,
```
Keshoid: **Minor — `events` should not be `pub`.** The `drain()` API is sufficient. A `pub Mutex<Vec<_>>` invites a future test to reach in and forget to lock (or hold the guard across `.await`). `pub(crate)` or fully private is enough.

src/node-control/service/src/audit/jsonl_writer.rs
```diff
@@ -122,15 +115,55 @@ impl AuditWriter {
...
+        AuditFileHeader {
+            schema_version: AUDIT_SCHEMA_VERSION,
+            service: "nodectl".into(),
+            service_version: env!("CARGO_PKG_VERSION").into(),
+            host: Self::hostname(),
+            started_at: Utc::now(),
+        }
+    }
+
+    fn hostname() -> String {
```
Keshoid: **Minor — `"unknown"` fallback is forensically annoying.** For an audit log, an event with `host: "unknown"` is a real loss of identity. At minimum emit a `tracing::warn!` once during writer init when the fallback is hit, so the operator can fix `$HOSTNAME` instead of discovering it only after the incident.

src/node-control/service/src/audit/event.rs
```diff
@@ -7,38 +7,230 @@
...
+    ) -> Self {
+        Self::new(
+            actor,
+            Self::node_target(node_id, election_id),
+            AuditOutcome::Success,
+            AuditEventPayload::ElectionsKeyGenerated { pubkey },
+        )
+    }
+
+    #[allow(clippy::too_many_arguments)]
```
Keshoid: **Minor — `#[allow(too_many_arguments)]` regresses the constructor's safety goal.**

Seven positional args, most of them `impl Into<String>` / `u64`. Easy to swap `election_id` with `submission_time` at a call site and the compiler can't catch it. The whole point of these typed constructors is to make construction safer than the previous struct-literal style — a positional 7-arg signature undoes part of that.

Consider a small params struct (`ElectionsStakeSubmittedParams { … }`) or a thin builder. Same idea applies as more producers (rewards / vault) get wired in SMA-99.5+.

src/node-control/service/src/audit/enums.rs
```diff
@@ -12,127 +12,201 @@ use serde::{Deserialize, Serialize};
 #[serde(tag = "event_type", content = "data", rename_all = "snake_case")]
 #[non_exhaustive]
 pub enum AuditEventPayload {
+    #[serde(rename = "elections.tick_failed")]
+    ElectionsTickFailed { reason: String },
```
Keshoid: I doubt this event will be useful for us. We have logs for tick failures.

src/node-control/service/src/audit/enums.rs
```diff
@@ -12,127 +12,201 @@ use serde::{Deserialize, Serialize};
...
         amount_nanotons: String,
         #[serde(skip_serializing_if = "Option::is_none")]
         tx_hash: Option<String>,
     },
 
+    #[serde(rename = "elections.withdraw_processed")]
+    ElectionsWithdrawProcessed { tx_hash: String },
+
+    #[serde(rename = "elections.withdraw_process_failed")]
+    ElectionsWithdrawProcessFailed { reason: String },
```
Keshoid: The name should be convenient with `ElectionsWithdrawProcessed`: `ElectionsWithdrawFailed`

src/node-control/service/src/audit/enums.rs
```diff
@@ -12,127 +12,201 @@ use serde::{Deserialize, Serialize};
...
     #[serde(rename = "rest_api.config_updated")]
-    RestApiConfigUpdated {
-        operation: String,
-        changes: serde_json::Value, // diff stays free-form
-    },
+    RestApiConfigUpdated { operation: String, changes: Vec<ConfigFieldChange> },
 
     #[serde(rename = "rest_api.auth_login_success")]
-    RestApiAuthLoginSuccess { username: String },
+    RestApiAuthLoginSuccess {},
```
Keshoid: Succeeded

src/node-control/service/src/audit/enums.rs
```diff
@@ -12,127 +12,201 @@ use serde::{Deserialize, Serialize};
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(rename_all = "snake_case")]
 #[non_exhaustive]
 pub enum StakeSkipReason {
     LowWalletBalance,
     WithdrawRequestsPending,
     PoolNotReady,
-    Other,
+    AdaptiveSleepPeriod,
+    AdaptiveWaitingPeriod,
```
Keshoid, line 112: Waiting vs Sleep - use one form in both. Waiting and Sleeping, for example.

src/node-control/service/src/audit/enums.rs
```diff
@@ -12,127 +12,201 @@ use serde::{Deserialize, Serialize};
...
+    NodeExcluded,
```
Keshoid: I would recommend to rename to ElectionsDisabled

src/node-control/service/src/elections/runner.rs
```diff
@@ -467,7 +528,18 @@ impl ElectionRunner {
...
                     self.refresh_validator_configs().await;
 
-                    if let Err(e) = &self.run().await {
+                    if let Err(e) = self.run().await {
+                        let election_id = tick_election_id.or({
+                            let id = self.past_elections_cache_id;
+                            (id > 0).then_some(id)
+                        });
+                        audit
+                            .record(AuditEvent::elections_tick_failed(
```
Keshoid: I think we don't need this event. We have logs for that.

Keshoid (changes requested):

src/node-control/service/src/elections/runner.rs
mrnkslv: fixed

src/node-control/service/src/elections/adaptive_strategy.rs
mrnkslv: done

src/node-control/service/src/elections/runner.rs
mrnkslv: done

src/node-control/service/src/elections/runner.rs
mrnkslv: done

src/node-control/service/src/elections/runner.rs
mrnkslv: done

src/node-control/service/src/elections/runner.rs
mrnkslv: outdated

src/node-control/service/src/elections/runner.rs
mrnkslv: done

src/node-control/service/src/audit/in_memory.rs
mrnkslv: accepted

src/node-control/service/src/audit/jsonl_writer.rs
mrnkslv: done

src/node-control/service/src/audit/event.rs
mrnkslv: done

src/node-control/service/src/audit/enums.rs
mrnkslv: deleted

src/node-control/service/src/audit/enums.rs
mrnkslv: done

src/node-control/service/src/audit/enums.rs
mrnkslv: renamed

src/node-control/service/src/audit/enums.rs
mrnkslv: accepted

src/node-control/service/src/elections/runner.rs
mrnkslv: deleted

src/node-control/service/src/audit/enums.rs
mrnkslv: renamed

Keshoid (approved):

PR #208, opened 2026-06-04
Feature/sma 104 audit log rest producers auditactorbuilder

mrnkslv:
## Summary

Add AuditActorBuilder and rest_audit helpers so record_client_ip and ip_anonymize apply only via rest_user. Emit config_updated from config handlers, login success/rejected from login, and token_rejected from auth middleware. Use fully qualified paths for third-party types in the audit module.

Tests cover PII policy, config diff emission, and auth audit events.

Closes SMA-104


src/node-control/service/src/elections/runner.rs
```diff
@@ -490,6 +490,9 @@ impl ElectionRunner {
             tokio::select! {
                 _ = interval.tick() => {
                     tracing::info!("TICK");
+                    let audit = self.audit.clone();
+                    let tick_election_id =
```
Keshoid: These two locals are never read anywhere in the tick block (the `audit` clone actually used by the tick is the separate one further down at ~line 858). Beyond being dead code (`unused_variables`), `tick_election_id` fires an extra `get_active_election_id()` RPC on every tick whose result is discarded.

This is also out of scope for SMA-104 (REST producers) — it looks like leaked groundwork for an elections-audit task. Please drop it from this branch or move it to the task it belongs to.

src/node-control/service/src/elections/adaptive_strategy.rs
```diff
@@ -90,6 +90,36 @@ pub(crate) fn adaptive_split50_status(
     None
 }
 
+/// When [`is_adaptive_split50_ready`] would return `false`, reports which wait gate blocked.
+pub(crate) fn adaptive_split50_defer_reason(
```
Keshoid: `adaptive_split50_defer_reason` has zero callers (grep finds only this definition) — dead code (`dead_code` warning). The `AdaptiveDeferReason` enum it returns is used elsewhere via `adaptive_split50_status`, but this standalone function is not. Like the runner.rs locals, it's unrelated to SMA-104; please remove it from this PR or move it to its owning task.

src/node-control/service/src/auth/middleware.rs
```diff
@@ -49,11 +52,21 @@ async fn require_role_impl(
     next: Next,
     min_role: Role,
 ) -> Response {
+    let headers = req.headers().clone();
```
Keshoid: This clones the entire `HeaderMap` on every authenticated request, but it's only consumed in the rejection arms — wasted on the success path (the common case). Since `req` isn't moved until `next.run(req)`, you can read `req.headers()` lazily inside each rejection arm instead and avoid the clone on the happy path.

Keshoid (commented):
Review for SMA-104 (audit log REST producers + AuditActorBuilder). Implementation matches the ticket well and the security-sensitive parts are handled carefully — all 9 named tests are present, no secrets/tokens/passwords reach payloads, and reading the PII policy live from `RuntimeConfigStore` (instead of a config snapshot, as the ticket suggested) is a good improvement.

Main concern: two unrelated, dead additions in the elections modules (flagged inline) that shouldn't ride along in this PR — one of them adds a wasted RPC per tick.

Minor coverage note (not tied to one line): only `elections.settings_updated` has a handler-level emission test. The other new emission sites (nodes/wallets/pools/bindings add+rm, ton_http_api, log, static_adnl, exclude/include) are untested end-to-end. A couple of entity-mutation assertions would lock in the `old`/`new` argument direction, which is easy to swap between the add and rm paths.

src/node-control/service/src/auth/middleware.rs
```diff
@@ -49,11 +52,21 @@ async fn require_role_impl(
...
+
     // Check live config: when auth is not configured, pass through.
     // This allows auth to be enabled/disabled at runtime via config reload.
     {
         let cfg = state.runtime_cfg.get();
         if cfg.http.auth.is_none() {
+            // Auth-disabled mode: inject a synthetic identity so mutation handlers
+            // can still require `Extension<Claims>` without enabling JWT checks.
+            req.extensions_mut().insert(Claims {
+                sub: "anonymous".into(),
```
Keshoid, line 67: Minor/informational: in auth-disabled mode the synthetic identity logs `role: min_role` — i.e. the route's *required* role, not a real grant — under `sub: "anonymous"`. That's a slightly misleading attribution in the audit trail. Fine to keep given there's no real identity, but worth being aware of.

src/node-control/service/src/http/rest_audit.rs
```diff
@@ -0,0 +1,214 @@
...
+
+pub async fn record_config_updated(
+    state: &AppState,
+    claims: &Claims,
+    headers: &axum::http::HeaderMap,
+    config_id: &str,
+    operation: &str,
+    changes: Vec<ConfigFieldChange>,
+) {
+    if changes.is_empty() {
```
Keshoid, line 65: Early-returning on an empty diff means a config update that produces no field-level change emits no audit event at all. Defensible, but it also means no-op update *attempts* aren't recorded. Please confirm that's intended vs. wanting a record-of-attempt.

src/node-control/service/src/http/config_handlers.rs
```diff
@@ -2192,6 +2370,27 @@ pub async fn v1_elections_settings_update_handler(
         })
         .map_err(map_try_update_save)?;
 
+    let after = state.runtime_cfg.get().elections.as_ref().expect("validated above").clone();
```
Keshoid: `.expect("validated above")` is an acceptable programming-invariant assertion, but it runs inside a request handler — if any `try_update_and_save` path can leave `elections = None`, this is a panic-as-DoS. Worth a quick confirmation that the update closure always sets `elections` to `Some` on the success path.

src/node-control/service/src/http/auth_tests.rs
```diff
@@ -170,13 +171,33 @@ async fn state_with_auth() -> AppState {
...
-        user_store: Arc::new(UserStore::new(rt as Arc<dyn RuntimeConfig>)),
+        user_store: Arc::new(UserStore::new(rt.clone() as Arc<dyn RuntimeConfig>)),
         login_rate_limiter: Arc::new(tokio::sync::Mutex::new(Default::default())),
         config_changed: Arc::new(tokio::sync::Notify::new()),
         audit: Arc::new(crate::audit::log::NoopAuditLog),
+        actor_builder: Arc::new(AuditActorBuilder::new(rt.clone())),
     }
 }
 
+async fn state_with_auth_audited() -> (AppState, Arc<InMemoryAuditLog>) {
```
Keshoid, line 141: Nit: `state_with_auth_audited` largely re-copies the `AppState` literal from `state_with_auth` (and `config_handlers_tests::state_audited` does the same). Consider a small shared builder to avoid the duplication drifting over time.

src/node-control/service/src/elections/runner.rs
mrnkslv: removed

src/node-control/service/src/elections/adaptive_strategy.rs
mrnkslv: duplicated when fixing conflicts. Removed

src/node-control/service/src/auth/middleware.rs
mrnkslv: done

src/node-control/service/src/auth/middleware.rs
mrnkslv: comment added

src/node-control/service/src/http/rest_audit.rs
mrnkslv: ✅

src/node-control/service/src/http/config_handlers.rs
mrnkslv: ok. returns 500 instead of panicking

src/node-control/service/src/http/auth_tests.rs
mrnkslv: done

Keshoid (approved):

PR #209, opened 2026-06-08
Feature/sma 105 audit log in memory ring buffer for read path

mrnkslv:
## Summary
Add AuditEventBuffer — fixed-capacity FIFO buffer (parking_lot::RwLock, never held across .await) that receives every audit event independently of the JSONL writer. JsonlAuditLog::record() pushes to the ring before the channel send, so events appear immediately and survive queue overflow.

AuditLogFactory now returns AuditComponents { log, ring } so the composition root can hand the ring to AppState.audit_ring for the REST read-path (GET /v1/elections) without touching disk on the hot path.

Default ring capacity: 100 events (configurable via ring_buffer_capacity). Lowered from 10_000 to bound in-memory cost on the hot path; also shrinks the in-ring dedup window for stake_skipped (JSONL writer queue unchanged).

Deduplication of repeated stake_skipped

AuditEvent::dedup_identity() returns a stable key for ElectionsStakeSkipped events (node_id, election_id, reason). JsonlAuditLog atomically checks and pushes via push_unless_dedup_duplicate(), preventing per-tick spam for persistent reasons (ElectionsDisabled, RecoverPending, etc.) while still recording on the first occurrence and again after a service restart (empty ring).

## Payload field renames
Remove the _nanotons suffix from all payload fields — the unit is implicit and the name "gram" may change soon:

stake_nanotons → stake amount_nanotons → amount required_nanotons → required available_nanotons → available total_nanotons → total

No in-repo consumer keys on the old field names; audit log not yet on main.

## Out of scope (carried in via merge)
src/node/tests/test_load_net/scripts/add-nominators-to-pool.ts — highload workchain default 0 + sendBatch chunking. Landed in f6a28f0 post-merge fixup alongside sma-104 renames; unrelated to SMA-105 ring buffer. Fixes singlehost load-driver batch deploy; no audit impact, CI green.

## Tests
ring_buffer.rs: push_below_capacity_keeps_all, push_at_capacity_evicts_oldest, snapshot_returns_in_order, filter_collect_only_matching_events, concurrent_push_and_snapshot_no_panic, zero_capacity_buffer_silently_drops, push_unless_dedup_duplicate_keeps_one_per_node_election_reason.

jsonl_log.rs: record_pushes_to_ring_even_when_queue_full, record_deduplicates_stake_skipped_by_node_election_and_reason.

299 tests pass.

Closes SMA-105.


src/node-control/service/src/audit/jsonl_log.rs
```diff
@@ -147,6 +158,18 @@ impl AuditLog for JsonlAuditLog {
...
+        // Prevents e.g. repeated elections.stake_skipped for the same (node, election, reason).
+        if let Some(key) = event.dedup_key()
+            && self.ring.contains_dedup_key(&key)
+        {
+            return;
+        }
+
+        // Push into ring first: readers see the event immediately and even
+        // queue-dropped events remain accessible on the REST read-path.
+        self.ring.push(event.clone());
```
Keshoid: Two points on the dedup path:

1. **No test covers dedup suppression (please add before merge).** `dedup_key()` + `contains_dedup_key()` are a headline deliverable of this PR, but nothing asserts the drop behaviour — `record_pushes_to_ring_even_when_queue_full` only tests ring capture. Suggest a test that records two `stake_skipped` events with the same `(node, election, reason)` and asserts the ring contains one, plus a third with a different reason that is kept.

2. **TOCTOU between the check and the push.** `contains_dedup_key` takes a read lock, releases it, then `push` takes a write lock. Two concurrent `record()` calls with the same key can both pass the check and both push. Benign for audit (worst case a rare duplicate line), but worth either a comment acknowledging it or folding the check into a single write-locked `push_dedup(key, event)`.

src/node-control/service/src/audit/ring_buffer.rs
```diff
@@ -0,0 +1,207 @@
...
+
+/// Fixed-capacity FIFO buffer of recent audit events for the REST read-path.
+///
+/// Populated on every [`crate::audit::log::AuditLog::record`] call — independent of the
+/// JSONL writer task. Events are available immediately and remain visible even when the
+/// writer queue is full.
+///
+/// `parking_lot::RwLock` is used intentionally: the critical section is purely in-memory
+/// (no IO, no `.await`), so a synchronous lock is correct and avoids the overhead of an
+/// async lock. The lock is **never held across an `.await` point**.
```
Keshoid: Nit on the rationale: writes (one `push` on every `record()`) dominate reads (only `GET /v1/elections`), so the "reads-dominate" case for `RwLock` doesn't actually hold here — a plain `Mutex` would be the textbook fit. The genuine justification (synchronous, never held across `.await`) applies equally to `Mutex`. No need to change the impl; just consider trimming the comment so it doesn't imply a read-heavy workload.

src/node-control/service/src/audit/ring_buffer.rs
```diff
@@ -0,0 +1,207 @@
...
+
+    pub fn capacity(&self) -> usize {
+        self.capacity
+    }
+
+    /// Returns `true` if the buffer already contains an event with the given
+    /// deduplication key. Used by [`crate::audit::jsonl_log::JsonlAuditLog`] to
+    /// suppress repeated identical `elections.stake_skipped` events within one election.
+    pub fn contains_dedup_key(&self, key: &str) -> bool {
+        self.inner.read().iter().any(|e| e.dedup_key().as_deref() == Some(key))
```
Keshoid: Minor: this scans the whole ring and calls `e.dedup_key()` — which allocates a `format!` string — for every entry, on every `record()` that has a dedup key. Negligible at capacity 100 and audit volume, but a structural compare (or `dedup_key()` returning a borrowed/cheaper key) would avoid up to N allocations per call.

src/node-control/common/src/app_config.rs
```diff
@@ -960,7 +960,7 @@ fn default_audit_include_payload() -> bool {
 }
 
 fn default_audit_ring_buffer_capacity() -> usize {
-    10_000
+    100
```
Keshoid, line 963: Confirm intent: the default ring capacity drops 100× (10_000 → 100). That shrinks both the REST read-path window and the dedup window. Likely fine for per-tick elections events, but it's a product-facing default change currently only surfaced via a test assertion — worth a line in the PR description.

src/node-control/commands/src/commands/nodectl/service_api_cmd.rs
```diff
@@ -751,9 +753,112 @@ fn print_elections_table(body: &str) -> anyhow::Result<()> {
...
+            let amount = data.get("amount").and_then(serde_json::Value::as_str).unwrap_or("-");
+            format!("amount={}", display_tons_from_str(amount))
+        }
+        "withdraw_processed" => {
+            let hash = data.get("msg_hash").and_then(serde_json::Value::as_str).unwrap_or("-");
+            format!("msg_hash={}", &hash[..hash.len().min(16)])
+        }
+        "key_generated" => {
+            let pubkey = data.get("pubkey").and_then(serde_json::Value::as_str).unwrap_or("-");
+            format!("pubkey={}…", &pubkey[..pubkey.len().min(12)])
```
Keshoid: Byte-index slicing (`&hash[..hash.len().min(16)]` and `&pubkey[..pubkey.len().min(12)]`) panics if the cut falls inside a multi-byte UTF-8 char. Safe today since these are hex/base64 ASCII, but `s.get(..n)` or `s.chars().take(n).collect()` would be panic-proof if a field ever carries non-ASCII.

src/node-control/service/src/audit/enums.rs
```diff
@@ -19,31 +19,26 @@ pub enum AuditEventPayload {
     },
 
     #[serde(rename = "elections.stake_submitted")]
-    ElectionsStakeSubmitted {
-        stake_nanotons: String,
-        max_factor: u32,
-        policy: String,
-        submission_time: u64,
-    },
+    ElectionsStakeSubmitted { stake: String, max_factor: u32, policy: String, submission_time: u64 },
```
Keshoid, line 22: The `*_nanotons` → bare-name renames (`stake`, `amount`, `required`, `available`, `total`) are a breaking change to the emitted JSONL event schema. Acknowledged in the PR description; just confirming no live consumer (Grafana/log parser/dashboard) is already keyed on the old field names before this ships.

src/node/tests/test_load_net/scripts/add-nominators-to-pool.ts
```diff
@@ -19,6 +19,8 @@
  *   MASTER_WALLET_ID (default 42)
  *   NOMINATOR_SUBWALLET_BASE (default 10000)
  *   NOMINATOR_WORKCHAIN (default **0** — required by pool; do not use -1 for nominators)
+ *   HIGHLOAD_WORKCHAIN (default **0** — must match nominators; orchestrator deploys nominator wallets in-batch on the same wc)
```
Keshoid, line 22: Scope: this load-driver change (highload workchain / batch chunking) is unrelated to the audit ring buffer — it rode in via the `f6a28f0` post-merge fixup commit. Harmless and CI-passing, but ideally it shouldn't be part of an SMA-105 PR. Consider splitting it out or calling it out in the description so reviewers aren't puzzled.

Keshoid (commented):
SMA-105 review — ring buffer for the audit read-path. The implementation is solid (correct push-before-send ordering, lock never held across `.await`, good concurrency test). Inline notes below: the only one I'd treat as a blocker is the missing dedup test; the rest are nits / confirm-intent.

src/node-control/service/src/audit/jsonl_log.rs
mrnkslv: done

src/node-control/service/src/audit/ring_buffer.rs
mrnkslv: done

src/node-control/service/src/audit/ring_buffer.rs
mrnkslv: fixed

src/node-control/common/src/app_config.rs
mrnkslv: added

mrnkslv: (Default ring capacity: 100 events (configurable via ring_buffer_capacity).)

src/node-control/commands/src/commands/nodectl/service_api_cmd.rs
mrnkslv: done

src/node-control/service/src/audit/enums.rs
mrnkslv: Confirmed: no in-repo consumer keys on the old `*_nanotons` field names.
`service_api_cmd`, serde golden tests, and elections projection all use the bare names
(`stake`, `required`, `available`, `amount`). Grafana/helm monitoring do not ingest
`audit.jsonl`.

src/node/tests/test_load_net/scripts/add-nominators-to-pool.ts
mrnkslv: added to description

Keshoid (approved):

PR #211, opened 2026-06-10
Feature/sma 106 audit log project events into get v1elections response

mrnkslv:
## Summary
Projects recent elections audit events from the in-memory ring buffer into the existing GET /v1/elections response, so operators see stake activity and errors in the same shape they already use — without a separate raw recent_events field.

The handler takes a single snapshot of elections-sourced events from the ring, builds a projection for up to three recent election cycles, and merges it into our_participants: stake_submissions are enriched from audit records, and last_error is derived from the latest stake skip, stake failure, or withdraw failure. The nodectl CLI elections table gains a Last error column aligned with the API.

## Changes
- Audit projection (audit/projection.rs)
Add project_elections() to fold ring-buffer events into per-election, per-node structures (stake_submissions, stake_skips, stake_failures, withdraws).
Add collect_recent_election_ids() — current cycle first, then other IDs seen in events (cap: 3).
Add merge_projection_into_participants() to merge projected data into live OurElectionParticipant snapshots without duplicate submissions.
Derive last_error from the chronologically latest error-class event (stake skipped / stake failed / withdraw failed).
Unit tests for grouping, filtering, merge behavior, and empty-ring no-op.
- GET /v1/elections handler (http_server_task.rs)
Read elections events once via audit_ring.filter_collect(AuditSource::Elections).
Project and merge into our_participants for the current election id.
Remove top-level recent_events from the response.
Handler tests: empty ring passthrough, stake submissions from audit, last_error from skip/failure/withdraw, end-to-end via JsonlAuditLog.
- CLI (service_api_cmd.rs)
Add Last error column to the elections participants table.
Remove client-side parsing/display of raw recent_events.

Closes SMA-106.


src/node-control/service/src/audit/jsonl_log.rs
```diff
@@ -153,13 +158,25 @@ impl AuditLog for JsonlAuditLog {
...
+        // Prevents e.g. repeated elections.stake_skipped for the same (node, election, reason).
+        if let Some(key) = event.dedup_key()
+            && self.ring.contains_dedup_key(&key)
+        {
+            return;
+        }
+
+        // Push into ring first: readers see the event immediately and even
+        // queue-dropped events remain accessible on the REST read-path.
+        self.ring.push(event.clone());
```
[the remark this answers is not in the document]
mrnkslv: ok

src/node-control/service/src/audit/ring_buffer.rs
```diff
@@ -0,0 +1,207 @@
...
+    pub fn capacity(&self) -> usize {
+        self.capacity
+    }
+
+    /// Returns `true` if the buffer already contains an event with the given
+    /// deduplication key. Used by [`crate::audit::jsonl_log::JsonlAuditLog`] to
+    /// suppress repeated identical `elections.stake_skipped` events within one election.
+    pub fn contains_dedup_key(&self, key: &str) -> bool {
+        self.inner.read().iter().any(|e| e.dedup_key().as_deref() == Some(key))
+    }
```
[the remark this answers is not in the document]
mrnkslv: fixed

src/node-control/service/src/audit/projection.rs
```diff
@@ -0,0 +1,414 @@
...
+    events: &[AuditEvent],
+    max: usize,
+) -> Vec<u64> {
+    let mut ids = Vec::new();
+    let mut seen = std::collections::HashSet::new();
+
+    if let Some(id) = current_election_id {
+        seen.insert(id);
+        ids.push(id);
+    }
```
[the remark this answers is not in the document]
mrnkslv: fixed

src/node-control/service/src/audit/projection.rs
```diff
@@ -0,0 +1,443 @@
...
+use chrono::{DateTime, Utc};
+use common::{
+    snapshot::{OurElectionParticipant, StakeSubmission},
+    time_format,
+    ton_utils::max_stake_factor_raw_to_multiplier,
+};
+use std::collections::BTreeMap;
+
+/// Projected stake submission from audit events.
+#[derive(Debug, Clone, PartialEq, Eq)]
```
Keshoid, line 22: The `Projected` prefix on these structs reads as noise — there are no competing variants, so it carries no disambiguating value. Suggest dropping it: `ProjectedStakeSkip → StakeSkip`, `ProjectedWithdraw → Withdraw`, `ProjectedStakeFailure → StakeFailure`.

`ProjectedStakeSubmission` is the only one with a real reason to differ: it collides with the imported `common::snapshot::StakeSubmission` used in this same file. But the projection ultimately produces exactly that snapshot type, so rather than keep a parallel struct we can store `snapshot::StakeSubmission` directly in `NodeElectionProjection.stake_submissions` and build it during `project_elections`. That removes the collision, the `projected_to_stake_submission` converter, and the fields that are currently written but never read (`ts`, `node_id`, `policy`, and the raw `max_factor`).

src/node-control/service/src/http/http_server_task.rs
```diff
@@ -496,26 +496,33 @@ pub async fn v1_elections_handler(
...
-        .filter_collect(|e| e.payload.source() == AuditSource::Elections)
-        .into_iter()
-        .rev()
-        .filter_map(|e| serde_json::to_value(e).ok())
-        .collect();
+    // Single snapshot of the ring; `project_elections` filters by `recent_ids` internally.
+    let elections_events =
+        state.audit_ring.filter_collect(|e| e.payload.source() == AuditSource::Elections);
+    let recent_ids = collect_recent_election_ids(current_election_id, &elections_events, 3);
+    let projection = project_elections(&elections_events, &recent_ids);
```
Keshoid: `elections_events` is traversed twice in a row here (`collect_recent_election_ids`, then `project_elections`). The first pass is actually dead work: `merge_projection_into_participants` is only ever called with `current_election_id`, and `our_participants` is a snapshot of the current cycle — so the two extra election ids collected here never match a participant and their projection is built for nothing.

Suggest dropping the `recent_ids` layer entirely and projecting the current cycle in a single pass:

```rust
let mut our_participants = view.our_participants;
if let Some(election_id) = current_election_id {
    let projection = project_elections(&elections_events, &[election_id]);
    merge_projection_into_participants(&mut our_participants, &projection, election_id);
}
```

This removes the `collect_recent_election_ids` call, the function itself (+ its unit tests), and its re-export in `mod.rs`. As a bonus, when `current_election_id` is `None` no projection is built at all.

If we do want to surface several cycles in the future, the better shape is a single fused function rather than two passes — `project_recent_elections(events, current_election_id, max)` that builds the full per-id projection in one pass over the events and then prunes to the most recent ids afterward. Since `election_id` is the (monotonic) election start timestamp, "most recent" is just the largest ids, so pruning is `retain` on the `BTreeMap` keys (highest `max` of them, current first) — no second traversal of the event slice, and `merge_*` would then loop over those ids instead of only the current one.

src/node-control/service/src/audit/projection.rs
```diff
@@ -0,0 +1,443 @@
...
+    for withdraw in &node_proj.withdraws {
+        if withdraw.outcome == AuditOutcome::Failure
+            && let Some(error) = &withdraw.error
+        {
+            candidates.push((withdraw.ts, format!("withdraw failed: {error}")));
+        }
+    }
+
+    candidates.sort_by_key(|(ts, _)| *ts);
+    candidates.last().map(|(_, msg)| msg.clone())
```
Keshoid: `latest_error_message` returns the most recent error-class event but never considers whether a *successful* `ElectionsStakeSubmitted` happened afterward. So a node that skips/fails at `t1` and then submits successfully at `t2 > t1` will still surface a `last_error` (and the CLI `Last error` column stays populated) even though the latest action succeeded.

If the field is meant as "last error ever seen", this is fine — but operators are likely to read it as "the current cycle is failing". Two options:
- Clear/suppress it when the latest successful submission for that node post-dates the latest error event, or
- Keep the behavior but make the historical semantics explicit (e.g. rename to `last_error_seen` / document it).

Worth a deliberate decision either way.

src/node-control/commands/src/commands/nodectl/service_api_cmd.rs
```diff
@@ -736,26 +737,28 @@ fn print_elections_table(body: &str) -> anyhow::Result<()> {
...
             submitted_at,
             max_factor,
+            last_error_display,
             pubkey,
             adnl,
         );
     }
     println!();
 
-    print_recent_events_table(&value);
```
Keshoid, line 757: Removing the `print_recent_events_table` call here is good, but two other call sites survive in the early-return branches of `print_elections_table` (the "No participants in response" and "No controlled participants" cases). Since the server now always sends an empty `recent_events` (skipped by `skip_serializing_if`), those two calls are permanent no-ops, and `print_recent_events_table` together with its helpers (`format_event_details`, `truncate_preview`) is effectively dead code.

Either remove the remaining two calls and the now-unused helpers here, or, if the full `recent_events` retirement is intentionally deferred to SMA-107 (#212 "final cleanup"), drop a note so it is not lost.

src/node-control/service/src/audit/projection.rs
mrnkslv: done

src/node-control/service/src/http/http_server_task.rs
mrnkslv: done

src/node-control/service/src/audit/projection.rs
mrnkslv: done

src/node-control/commands/src/commands/nodectl/service_api_cmd.rs
mrnkslv: fixed

Keshoid (approved):

